PR #43, opened 2015-02-14
Support Dart VM >= 1.9.0 websocket observatory protocol.

cbracken:
In Dart SDK 1.9.0, access to the VM Observatory via the Chromium remote debugging protocol is no longer supported, and replaced with a JSON-rpc websocket interface via the observatory port. This behaviour is consistent with the VM websocket interface.

This patch applies the old behaviour on pre-1.9.0 VMs and the websocket JSON-rpc protocol for 1.9.0 and later.


cbracken:
@srawlins

Issue #31, opened 2014-07-31
update README.md to explain how to use `pub global` to run the script


zoechi:
Isn't this a perfect candidate to use the new `pub global` feature to run the script?

```
pub global activate coverage
pub global run coverage:collect_coverage --port=NNNN -o coverage.json --resume-isolates
```
I assume `pub global` has not yet landed in stable, but can't be long ;-)

cbracken:
Agreed - this makes perfect sense. I think it makes sense to land #36 done first, then follow up with this request.

cbracken:
Readme is updated via your patch c5d8f7caccdfa6c3b8aa8c520749c144572ba7e5. Thanks!

PR #54, opened 2015-03-14
#31 add support for `pub global run`

zoechi:


devoncarew:
@cbracken

cbracken:
lgtm

seaneagan:
I'm a little late, but it seems like it would be better to namespace it to `dart` e.g. `dartcoverage format` `dartcoverage collect`.  `collect_coverage` is probably not likely to conflict, but it is unclear when you see it in a script what it does.

cbracken:
@seaneagan There's work in progress to extract out a reasonable coverage API and replace both these scripts with something a bit more straightforward. I think that's excellent advice. Filed #58.

devoncarew:
Hmm, what about prefixing with the package name? I think having the executable names namespaced by the package their contained in makes sense. So, in this case, something like `coverage_collect` and `coverage_format`.

cbracken:
I believe what Sean is suggesting is that it would be useful to clarify in the name that this is a Dart-specific coverage tool and not C/Java/Intercal/etc.

seaneagan:
@cbracken yes, that was my reason for wanting to prefix with `dart`.  I like `dartcov`  That might also make sense as the package name, similar to `dartdoc` and `dart_style`.  The `linter` was renamed from `dart_lint`, so it doesn't match that.

cbracken:
I like `dartcov`; added a note to #58

PR #56, opened 2015-03-14
Task53 exception on internal

zoechi:


cbracken:
Closing since the main issue (introduced in 3d30f31446be5d4bb80e48f42e81a62371c97bcd) was resolved by 2d4cc9374165d4f503244ab703c997811f264ffe. If you've seen it throw since then, please re-open.

You should be able to use with the `test` package. The unit tests for this package are written using it and I have no problem collecting coverage against our own unit tests. Something like the following should work:

```
$ dart --observe=9090 test/collect_coverage_test.dart
$ dart bin/collect_coverage.dart --port 9090 -r > coverage.json
```

PR #66, opened 2015-05-04
Warn on missing file instead of failing.

ekweible:
If the resolver can't resolve a file path, it fails with a FileSystemException. I've run into scenarios when running coverage on tests in dartium where a Dart SDK path is malformed, and warning on these failed paths instead of failing altogether would be a better solution in my opinion.


cbracken:
Resolved via: 2d4cc9374165d4f503244ab703c997811f264ffe

PR #67, opened 2015-05-04
Add option to restrict coverage breadth.

ekweible:
Currently there is no way to specify which files you actually want coverage on during the `collect_coverage` step. As a workaround, I've added a `report-on` option to the `format_coverage` that takes one or many paths specifying the files/directories for which coverage should actually be formatted and included in the final lcov report.

Example usage:
```
pub run coverage:format_coverage -l --package-root=packages -i coverage.json -o coverage.lcov --report-on=lib/
```

The `coverage.lcov` output will then only include coverage data for files within the `lib/` directory.

[Review on Reviewable](https://reviewable.io/reviews/dart-lang/coverage/67)

### DIFF
--- a/lib/src/formatter.dart
+++ b/lib/src/formatter.dart
@@ -1,6 +1,7 @@
 library coverage.formatter;
 
 import 'dart:async';
+import 'dart:io';
 
 import 'resolver.dart';
 
@@ -17,12 +18,15 @@ class LcovFormatter implements Formatter {
   final Resolver resolver;
   LcovFormatter(this.resolver);
 
-  Future<String> format(Map hitmap) async {
+  Future<String> format(Map hitmap, {List<String> reportOn}) async {
     var buf = new StringBuffer();
+    var reportOnPaths = reportOn != null ? reportOn.map((path) => new File(path).absolute.path) : [];
 
     hitmap.forEach((key, v) {
       var source = resolver.resolve(key);
       if (source == null) {
+        return;
+      } else if (!reportOnPaths.isEmpty && !reportOnPaths.any((p) => source.startsWith(p))) {
         return;
       }
       buf.write('SF:${source}\n');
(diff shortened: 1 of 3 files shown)


kevmoo:
@ekweible would you rebase this on latest?

@cbracken Is this a solid change?

ekweible:
@kevmoo I merged master and resolved conflicts. @cbracken could you comment on if this is a change you would like to integrate? Also it looks like the Travis build failed for the dev sdk because of a transient pub get timeout, could you re-run that build?

ekweible:
@kevmoo @cbracken any update on this?

kevmoo:
1. Please rebase so there is one commit
2. Update `Formatter.format` to include the new argument
3. Please add tests! `test/lcov_test.dart` is a good spot.

kevmoo:
er, on 1 above: make sure your commit is a clean application on top of the current master

lib/src/formatter.dart
```diff
@@ -17,13 +18,16 @@ class LcovFormatter implements Formatter {
...
-  Future<String> format(Map hitmap) async {
+  Future<String> format(Map hitmap, {List<String> reportOn}) async {
     var buf = new StringBuffer();
-
+    var reportOnPaths = reportOn != null ? reportOn.map((path) => new File(path).absolute.path) : [];
     hitmap.forEach((key, v) {
       var source = resolver.resolve(key);
       if (source == null) {
         return;
+      } else if (reportOn != null && reportOn.length > 0 &&
```
kevmoo: you know `reportsOnPaths` exists. Just use that. No null or length check.

lib/src/formatter.dart
```diff
@@ -17,13 +18,16 @@ class LcovFormatter implements Formatter {
...
+                 reportOnPaths.where((p) => source.startsWith(p)).length == 0) {
```
kevmoo: Instead of `where` use `!reportsOnPaths.any((p) => ...)`

You calculate the whole length of the enumerable, which can be avoided.

ekweible:
@kevmoo thanks for the feedback! Had to keep the `isEmpty` check because it indicates that the `report-on` option isn't being used, otherwise it wouldn't produce any coverage at all.

I've also rebased on the latest master.

ekweible:
@cbracken any word on if this is a reasonable change that will be accepted?

kevmoo:
@ekweible Thanks for your patience. I'll follow-up with some cleanup. It's good to update the changelog and the pubspec.yaml version. See next CL.

PR #82, opened 2015-05-22
Migrate resolver to async-await syntax.

cbracken:


kevmoo:
oops...

cbracken:
Fixed. +1 for running assertions on :)

kevmoo:
See https://github.com/dart-lang/coverage/pull/85

cbracken:
Even better than #85.

lib/src/resolver.dart
```diff
@@ -84,22 +84,11 @@ class Loader {
...
-          .catchError((e) {
-        failed.add(uri);
-        c.complete(null);
-      });
-      return c.future;
+  Future<List<String>> load(String path) async {
+    try {
+      return new File(path).readAsLines();
+    } catch (_) {
+      return null;
```
kevmoo: do we still care about populating `failed`?

kevmoo:
Love it. Just one nit.

lib/src/resolver.dart
cbracken: Yep. Done.

kevmoo:
ship it

PR #83, opened 2015-05-22
Migrate doRetry() to async-await syntax.

cbracken:


kevmoo:
See https://github.com/dart-lang/coverage/pull/84

PR #93, opened 2015-06-04
Final tweaks before 0.7.0 release

kevmoo:
Cool?

[Review on Reviewable](https://reviewable.io/reviews/dart-lang/coverage/93)

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -1,0 +1,69 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library coverage.collect;
+
+import 'dart:async';
+
+import 'devtools.dart';
+import 'util.dart';
+
+const _retryInterval = const Duration(milliseconds: 200);
+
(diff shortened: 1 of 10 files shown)


cbracken:
Very nice! Thanks for extracting getAllCoverage(); this helps a ton with the PR I'm working on.

lgtm

---

Review status: 0 of 10 files reviewed, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dart-lang/coverage/93)

lib/src/collect.dart
```diff
@@ -0,0 +1,58 @@
...
+library coverage.collect;
+
+import 'dart:async';
+
+import 'devtools.dart';
+import 'util.dart';
+
+const _retryInterval = const Duration(milliseconds: 200);
+
+Future<Map> collect(String host, int port, bool resume, bool waitPaused,
```
zoechi: Would be great if you could expose `collect` as public to be used by other packages. This allows to collect coverage in-process just with a method call (without launching another Dart process for collect_coveralls and reading stdout). I can use it this way but it is against the guidelines to import from otherpackage/src.

kevmoo: Ooo! Yes.

Create a pull request which just exports this guy from lib/coverage.dart – is easy

Issue #39, opened 2015-02-02
Explain how coverage statistics are gathered.


bgourlie:
It would be nice if it were explained how coverage statistics were gathered.  As far as I know, observatory plays some part in it, but that's about it.  Ultimately, I'm trying to figure out why my package is showing zero test coverage.  Knowing how this package works would be a good start to figuring that out.

Anyhow, I have an issue on my tracker here https://github.com/bgourlie/noBS/issues/3 if anyone would like to take a look at my specific setup.

cbracken:
There are two components to the coverage package:
   1. `bin/collect_coverage.dart`: connects to a running VM or Dartium instance and, for each isolates gathers a hitmap (map of line number to hitcount) for each script loaded by that isolate. These hitmaps are then emitted to stdout in JSON format. Alternatively, the same JSON can be generated by running the VM with `--coverage_dir=somedir`.
   1. `bin/format_coverage.dart`: takes the JSON generated by either method in the step above, merges it down to a single set of hitmaps, one per script file, and emits it in the specified format: either pretty-printed source marked up with a per-line hitcount or in LCOV format.

Last I looked, Coveralls relies on `--coverage_dir` which only works for VM coverage. I'm looking at options for automating collection of both VM and web coverage (which will likely require Dartium content-shell to be installed).  #48 tracks that work.

I'll add documentation to README.md.

PR #94, opened 2015-06-05
expose lib/src/collect

zoechi:
I had the impression `true` for `wait` should work here https://github.com/dart-lang/coverage/compare/master...bwu-dart-contributing:master#diff-43df7f33c6ee2745896deffe2f16bfb4R77 but it didn't. Maybe I just don't understand how this is supposed to work.

I created a new test file even though a lot of code is just duplicated.
I thought it would be too messy if I put it in the `collect_coverage_test` file as well.

Some tests in `collect_coverage_test.dart` also the test I added only works half of the time on my machine even with much higher timeout. 
I wasted half of this day trying to make this work in a package of mine (didn't work reliable with the code before or after #39)

testout

The attached file is the test output (actually a .txt file but GitHub won't let me attach it this way).

### DIFF
--- a/lib/coverage.dart
+++ b/lib/coverage.dart
@@ -3,3 +3,4 @@ library coverage;
 export 'src/hitmap.dart';
 export 'src/formatter.dart';
 export 'src/resolver.dart';
+export 'src/collect.dart';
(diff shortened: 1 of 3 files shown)


cbracken:
lgtm. 

For the CLA, can I get you to confirm just for the record that @zoechigist is also you and that you're okay with merging?

zoechi:
Yes that's me. Seems to be a problem GitHub has because I renamed my account in the early days.

cbracken:
Thanks! Merged.

Issue #36, opened 2014-09-15
Single command to generate coverage


sethladd:
Right now, it's a two step process to spin up a VM, then run a separate command, then convert the output.

Can we make it a single command?

cbracken:
Closing as duplicate of #9.

PR #95, opened 2015-06-08
Adding runAndCollect

kevmoo:
This is progress towards https://github.com/dart-lang/coverage/issues/36

Just need to add in the entry point the bin

[Review on Reviewable](https://reviewable.io/reviews/dart-lang/coverage/95)

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -33,8 +33,8 @@ Future<Map> _getAllCoverage(VMService service) async {
   var vm = await service.getVM();
   var allCoverage = [];
 
-  for (var isolate in vm.isolates) {
-    var coverage = await service.getCoverage(isolate.id);
+  for (var isolateRef in vm.isolates) {
+    var coverage = await service.getCoverage(isolateRef.id);
     allCoverage.addAll(coverage.coverage);
   }
   return {'type': 'CodeCoverage', 'coverage': allCoverage};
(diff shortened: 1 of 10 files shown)


lib/src/util.dart
```diff
@@ -44,3 +45,22 @@ Future retry(Future f(), Duration interval, {Duration timeout}) async {
...
+
+/// Returns an open port by creating a temporary Socket
+Future<int> getOpenPort() async {
+  ServerSocket socket;
+
+  try {
+    socket = await ServerSocket.bind(InternetAddress.LOOPBACK_IP_V4, 0);
+  } catch (_) {
+    // try again v/ V6 only. Slight possibility that V4 is disabled
+    socket = await ServerSocket.bind(InternetAddress.LOOPBACK_IP_V6, 0,
```
zoechi: Doesn't `bind` try both IPv6 and IPv4 when `v60only` is `false`?

kevmoo: I'm being paranoid. There are still systems that have IPv6 disabled. More likely a system has IPv4 and not v6 then the other way around.

I'm not 100% sure what happens on a system that is v4-only w/ the V6 API.

zoechi: Same here, I hoped you can clarify ;-)

kevmoo: I don't have a machine w/ v6 disabled, so it's tough to figure out. :smile_cat:

kevmoo:
@cbracken I have cookies... :cookie:

cbracken:
I like cookies but I've got a few P0s first. Will get to this by EOW, where EOW is defined as Saturday morning midnight PDT. ;)

kevmoo:
@ekweible @srawlins Could I get a look-through by one of you guys? Look okay?

lib/src/collect.dart
```diff
@@ -42,17 +42,18 @@ Future<Map> _getAllCoverage(VMService service) async {
 
 Future _resumeIsolates(VMService service) async {
   var vm = await service.getVM();
-  var isolateRequests = vm.isolates.map((i) => service.resume(i.id));
-  return Future.wait(isolateRequests);
+  for (var isolate in vm.isolates) {
```
ekweible: Could we call this `isolateRef`as well to be consistent?

ekweible:
+1 looks good to me.

lib/src/collect.dart
kevmoo: Good point. Fixed

PR #105, opened 2016-02-16
Updates

kevmoo:
Tests were failing with the latest SDK

I fixed that and updated our infrastructure to use vm_service_client


kevmoo:
@cbracken This would be a good start on the migration :smile:

evanweible-wf:
@kevmoo @cbracken any update on this? Would be great to have this fix!

kevmoo:
I'll ask @cbracken to review after the fact so we can move forward. :smile:

johnbland-wf:
@kevmoo how long do you think until this is tagged?

kevmoo:
@johnbland-wf nowish :smile:

johnbland-wf:
Yeah, saw that after commenting. lol. :) Still fails for me, but I'll double check some other things here first.

PR #108, opened 2016-05-26
Support for latest Dart VM service API

kevmoo:
This just supports outputting lcov on a running process

But it gets us pretty far!


cbracken:
Thanks for all the work on this, but before we spend the time reviewing what is effectively a rewrite with a lot of breaking changes, please do an internal pull to verify this works for the internal codebase. Given how substantial the rewrite is, I'd rather we knew this worked on a large real-world codebase before spending the review time.

We've had coverage for 1.16+ SDKs working internally since late April, but haven't been able to push because all the changes being made on github haven't been being pulled in as they went.

lib/src/resolver.dart
```diff
@@ -1,87 +1,61 @@
...
-    return null;
-  }
 
-  /// Returns a canonicalized path, or `null` if the path cannot be resolved.
-  String resolveSymbolicLinks(String path) {
-    var normalizedPath = p.normalize(path);
-    var type = FileSystemEntity.typeSync(normalizedPath, followLinks: true);
-    if (type == FileSystemEntityType.NOT_FOUND) return null;
-    return new File(normalizedPath).resolveSymbolicLinksSync();
+    throw new Exception('Could not resolve "$uri"');
```
DisDis: Need return null. As in the case of 'http://localhost: 8089/all_test.dart' exception occurs.

kevmoo: @DisDis I don't have issues running `test/test_all.dart`.

What are you referring to here?

DisDis: @kevmoo 
i'm execute dartium and collect coverage info. The previous version(7.3) of collecting coverage information without error, because it can handle the url. Example http://localhost:8080/my_browser_test.dart and etc.
Reproduce steps:
1) `pub serve test`
2) `pub run test --concurrency=1 --pub-serve=8080 -p dartium`
3) wait of the end of the tests. I used special test package, it is not kill isolations and dartium. So I can collect all the data coverage.
4) `pub global run coverage collect --port=<REMOTE_PORT> --host=localhost -o coverage.lcov`
 Data collection falls because the method `String resolve(Uri uri) [lib/src/resolver.dart]` comes url = `'http://localhost:8080/my_browser_test.dart'`

PR #111, opened 2016-06-03
Temporarily disable coveralls.io

cbracken:
The coveralls package relies on the (now eliminted) VM --coverage-dir
flag and needs to be updated to work again, or an alternative developed.


cbracken:
Dropping. No harm in leaving it on for now. Will look at updating the coveralls package.

PR #125, opened 2016-06-23
Collect packages from packageRoot/.packages in Resolver

Anrock:
Hello.
I'm not sure if this a bug in `coverage` or something wrong in my project, but `format_coverage` can't resolve `package:` uris for my project, passed as package-root, because `Resolver::_packages` aren't initialized if `--package-root` option is given.

I've made a patch, that makes Resolver parse `<package-root>/.packages` file if `packageRoot` is specified. Now `format_coverage` outputs coverage info for my project.

### DIFF
--- a/lib/src/resolver.dart
+++ b/lib/src/resolver.dart
@@ -14,8 +14,9 @@ class Resolver {
   final List<String> failed = [];
   Map<String, Uri> _packages;
 
-  Resolver({String packagesPath, this.packageRoot, this.sdkRoot})
-      : packagesPath = packagesPath,
+  Resolver({String packagesPath, String packageRoot, this.sdkRoot})
+      : packageRoot = packageRoot,
+        packagesPath = packagesPath != null ? packagesPath : packageRoot + "/.packages",
         _packages = packagesPath != null ? _parsePackages(packagesPath) : null;
 
   /// Returns the absolute path wrt. to the given environment or null, if the


Anrock:
Oh, unit-tests failed, but you get the idea.

cbracken:
Can you provide repro instructions (just the command line you're invoking) + a stack trace? I'm able to successfully use either `--packages` or `--package-root` with 0.7.8+2.

`_packages` is expected to be null if you use `--package-root`, it's just a local cache of the `.packages` file, which only applies when using `.packages` files via the `--packages` option. When `_packages` is null, we skip the conditional block and jump directly to [line 68](https://github.com/dart-lang/coverage/blob/master/lib/src/resolver.dart#L68) which resolves the source file against the package-root you've specified. 

Based on your patch, it looks like you might be specifying the package directory, instead of a package-root. The `--package-root` option on format_coverage (and on dart2js, and the vm) specifies the directory relative to which `package:` imports are resolved. If you're using pub in your workflow, that would typically be the `packages` directory in the package in question -- i.e. a sibling of the `pubspec.yaml`.

Example invocation:
```
$ dart bin/format_coverage.dart --package-root ../source_span/packages -i /tmp/coverage.json
/Users/cbracken/src/source_span/test/file_message_test.dart
       |// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
       |// for details. All rights reserved. Use of this source code is governed by a
```

And the same with `--lcov`.

Anrock:
On coverage-0.7.9 (it was 0.7.7 yesterday).
Running `pub run test -p content-shell --pause-after-load` and `collect_coverage -p N -r` to collect coverage.
This produces coverage.json where files from my project are mentioned in form `package:<package_name>/path/to/file.dart`
Then running `format_coverage -i coverage.json -o lcov --package-root ~/Desktop/project` produces empty `lcov` file. `~/Desktop/project` at it's root contains `pubspec.yaml` and `.packages` file with line `<package name>:lib/`.
When 68 line is executed, value of `p.join(packageRoot,uri.path)` is equal to `~/Desktop/project/<package_name>/path/to/file.dart` which of course doesn't exist. It's actual location is `~/Desktop/project/lib/path/to/file.dart`.

cbracken:
Again -- you need to specify the pacakge-root, *not* the package directory. The package-root for code executed in a given package foo is typically at `foo/packages`.

If you're using pub, and based on a guess from the command-line you've given, I suspect you want:
```
format_coverage -l -i coverage.json -o lcov --package-root ~/Desktop/my_project/packages
```

Note the `/packages` after `my_project`.

Anrock:
That works, thank you. Sorry for wasted time.

cbracken:
No problem! As I say, I'll update the README to make things a bit clearer.

As I say, the flag name is pretty confusing. On the plus side, the Dart ecosystem is moving to `.packages` files as the primary means of resolving packages. It's got the advantage of being more flexible and avoids the need for the whole symlink farm pub generates. In that case,  your command-line would be:
```
format_coverage -l -i coverage.json -o lcov --packages ~/Desktop/my_project/.packages
```

which is a bit less confusingly named.

PR #126, opened 2016-06-23
Add a basePath argument to the LCOV formatter

abarth:
The resulting LCOV file will use paths relative to this base path.


cbracken:
lgtm!

PR #127, opened 2016-06-23
`format_coverage`: add `--base-directory` option.

cbracken:
LCOV output script paths are emitted relative to this directory if specified or
as absolute paths, if not.

### DIFF
--- a/bin/format_coverage.dart
+++ b/bin/format_coverage.dart
@@ -13,6 +13,7 @@ class Environment {
   String sdkRoot;
   String pkgRoot;
   String packagesPath;
+  String baseDirectory;
   String input;
   IOSink output;
   List<String> reportOn;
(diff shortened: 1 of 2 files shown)


cbracken:
@abarth any preference for `--base-directory` vs `--base-path`? Went with 'directory' to match `lcov`'s option, but 'path' has the advantage of being shorter.

abarth:
Whatever you think is best.  I went with "path" in the Dart code to emphasize that it's a String rather than a Directory object.  On the command line, i doubt it maters.  You probably want to `p.absolute` the value somewhere in your patch.

bin/format_coverage.dart
```diff
@@ -186,6 +190,8 @@ Environment parseArgs(List<String> arguments) {
     stderr.writeln('warning: ignoring --bazel-workspace: --bazel not set');
   }
 
+  env.baseDirectory = args['base-directory'];
```
abarth: I'd call `p.absolute` here.

abarth:
LGTM

bin/format_coverage.dart
cbracken: Done.

PR #130, opened 2016-08-03
Extract format reportOn and basePath parameters to constructor

cbracken:
This allows pre-configured Formatter objects to be passed to APIs.


abarth:
LGTM

PR #135, opened 2016-10-17
Migrate tests from packages dir to package spec

cbracken:
As of Dart SDK 1.20.0, pub no longer creates a packages directory by
default, relying instead on package spec (.packages) files.


cbracken:
@tvolkert

tvolkert:
LGTM

PR #150, opened 2017-01-04
Add Evan Weible to authors

cbracken:


ekweible:
@cbracken LGTM, thanks!

PR #154, opened 2017-01-05
Default to checked mode in runAndCollect

cbracken:
In almost all cases, developers should be defaulting to checked mode
when running tests. This change brings behaviour in line with the
package:test default.


cbracken:
Fixes #154

Issue #158, opened 2017-01-18
Version 0.9.0 broke JetBrains coverage collection


daniel-v:
With the introduction of `--uri` parameter over port, JetBrains' Dart plugin can no longer collect the coverage data. In fact it would not even start.

Relevant lines in their repo:

`https://github.com/JetBrains/intellij-plugins/blob/fcb89e045f291e5117012ba260182a1142fe75b8/Dart/src/com/jetbrains/lang/dart/coverage/DartCoverageProgramRunner.java#L92`

Related JetBrains issue:
https://youtrack.jetbrains.com/issue/WEB-25047

Would it not make sense to deprecate port and host options first and remove them at later time? Give the 3rd parties a bit of time to adjust.

cbracken:
Thanks for reporting! 

According to the [breaking change notes](https://groups.google.com/a/dartlang.org/forum/#!msg/announce/VxSw-V5tx8k/wPV0GfX7BwAJ) for Dart VM 1.22, the auth token will be required by default when it goes stable, so the `--host` and `--port` options won't be enough. 

That said, I've spoken to @johnmccutchan and it sounds like they're going to put that off a bit if possible. In the meantime, I'll bring back `--host` and `--port`.

cbracken:
Fixed by #159 and released as version 0.9.1.

daniel-v:
@cbracken thanks for the quick solution and response!

cbracken:
Anytime! Thanks for filing the issue. In all likelihood, I'll keep these options around until (a) IntelliJ is patched to deal with URIs, and (b) the VM flips its default to require the auth token, plus one stable release.

The IntelliJ patch should definitely happen sooner rather than later, just to be safe. @devoncarew can you or @jwren follow up on that?

alexander-doroshko:
On it

PR #159, opened 2017-01-18
Add back support for --host and --port options

cbracken:
This is a temporary measure for backward compatibility until dependent
code has migrated to the --uri option.


cbracken:
Fixes #158.

Note: this is a partial revert that adds back the `--host` and `--port` command line options only. I does not add back the previous APIs supporting `host` and `port` parameters. Those who need the old APIs should pin their dependency to `coverage: >=0.8.0 <0.9.0`.

PR #178, opened 2017-03-04
Enable lint: avoid_as

cbracken:


zoechi:
I thought avoiding `as` doesn't provide any value anymore.

cbracken:
In strong mode, you're right. The value in avoiding it here is more about the readability of the code by being more explicit in the declared types, which obviates the need for `as`. Use of `as` implies a need to cast because the type differs from what's declared; with the more precise typing, that's no longer the case.

PR #185, opened 2017-04-27
Reduce memory consumption during coverage collection

joeconwaystk:
This PR looks to address challenges faced by collecting coverage on tests with a significant amount of isolate spawning. See also #184.

1. Adds optional StringSink argument to collection to allow for progress output. This is key for long-running tasks in CI that may be shutdown if no output is received for a period of time.
2. Resumes isolates immediately after their coverage has been collected. Previous strategy waited until all isolates were collected, which prevented their allocations from being released and ballooned memory consumption of to unsustainable levels on CI boxes.

This is my first jump into this package, so if I am misunderstanding something, will make the appropriate fix.

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -39,9 +39,11 @@ Future<Map<String, dynamic>> collect(
     return await _getAllCoverage(vmService, outputSink: outputSink);
   } finally {
     if (resume) {
-      await _resumeIsolates(vmService);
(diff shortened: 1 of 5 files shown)


cbracken:
Thanks for the issue and the PR, and apologies for the slow response! Haven't reviewed yet, but will grab a look over the weekend.

PR #190, opened 2017-10-02
Increase SDK lower bound to >=1.21.0

cbracken:
Support for method generics, which pacakge:coverage users, was
introduced in Dart SDK 1.21.0. Release notes:
https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md#1210---2016-12-07


cbracken:
Tested and verified on 1.21.0 with a `pub get downgrade`.

Issue #196, opened 2017-10-25
Async Function "Miss" Closing Brace


zacharynewton-wk:
When running coverage on an async function in dart, the closing function brace is always considered a "miss".
![image](https://user-images.githubusercontent.com/28708301/32008659-6311dc76-b972-11e7-9692-66cdfde0c516.png)
Where would be a good starting point to helping get this fixed?

cbracken:
@zacharynewton-wk do you know if this is this something that started recently or has it been an issue for a while? My guess is that the issue will be either be with how the source is tokenized for the purposes of hit count in the VM service or with the mapping of those source tokens back to the lines in the original script.

/cc @rmacnak-google for thoughts

zacharynewton-wk:
@cbracken this issue has been existing for as long as I started using dart and codecov at Workiva in May 2017

rmacnak-google:
If I look at an async function in Observatory, I can see there's a synthetic call site at the closing brace for "competeError", and it goes green if I throw inside the function. I'm not sure why this call site shows up when the "complete" one doesn't. I'd consider this a VM bug.

cbracken:
Thanks! I'll re-file this against dart-lang/sdk but keep this one open. Once it's fixed, I'll add a test case here.

zacharynewton-wk:
Awesome! Thanks @rmacnak-google and @cbracken

cbracken:
For ref: the fix from @rmacnak-google: https://dart-review.googlesource.com/c/sdk/+/16605

cbracken:
This has rolled out in Dart SDK 2.0.0-dev.5.0. #201 updates the tests.

Note that Dart SDK 2.0.0-dev.6.0 includes an additional [coverage-related fix]](https://dart-review.googlesource.com/c/sdk/+/16684/2) that ensures that function definitions are marked coverable. Example:
```dart
int get meaningOfLifeTheUniverseAndEverything = 42;
```

Tests for that

cbracken:
Closing as resolved now that this is rolled out. 

@zacharynewton-wk Thanks for reporting!

PR #198, opened 2017-10-25
Test coverage for synchronous function call

cbracken:
Adds a test to verify expected coverage is generated for a synchronous
function with two branches, only one of which is executed.


cbracken:
Additional testing prompted by #196.

This will be followed by a test for an async function call momentarily.

PR #199, opened 2017-10-25
Test coverage for async function call

cbracken:
Adds a test to verify expected coverage is generated for an asynchronous
function with two branches, only one of which is executed.

Line 21 is currently marked coverable and not hit due to a VM bug. Once
https://github.com/dart-lang/sdk/issues/31222 is fixed and the new VM
released to stable, we should remove the test expectation for that line
and close https://github.com/dart-lang/coverage/issues/196.


cbracken:
Adds preliminary test coverage for the issue uncovered in #196. The fix for that issue has now landed on master in the SDK. Once the SDK is released to stable channel, we'll resolve the TODO.

PR #201, opened 2017-11-02
Closing braces on async functions are not coverable

cbracken:
This eliminates test coverage expectations for the closing brace of
async functions, which was previously emitted as coverable by the VM
service (it shimmed in completeError as a hidden token on the closing
brace). See: https://github.com/dart-lang/coverage/issues/196

As of https://dart-review.googlesource.com/c/sdk/+/16605, that issue is
resolved in the VM. This patch eliminates test code designed to break
when that change rolls out.

### DIFF
--- a/test/collect_coverage_test.dart
+++ b/test/collect_coverage_test.dart
@@ -57,23 +57,27 @@ void main() {
     expect(hitMap, contains(_sampleAppFileUri));
 
     Map<int, int> isolateFile = hitMap[_isolateLibFileUri];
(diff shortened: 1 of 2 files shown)


cbracken:
Note that this will pass on Dart VM version 2.0.0-dev.5.0, but fail on Dart SDK 2.0.0-dev.6.0 due to a [second change](https://dart-review.googlesource.com/c/sdk/+/16684/2) to mark the function definition as a coverage tick as well. That will be handled in a followup PR.

PR #205, opened 2017-11-02
Bugfix in tests_all: coverage_api test was not being run

cbracken:
coverage_test was being run twice.


cbracken:
Travis always runs via `pub run test`, so all tests were being run in CI, but `tests_all.dart` is used for coverage collection on the coverage package itself.

PR #209, opened 2017-11-08
Fix warnings related to fuzzy arrows.

munificent:
These will become an arrow in Dart 2.0. See this bug for more context:

  https://github.com/dart-lang/sdk/issues/29630

I also fixed a couple of unused variable warnings while I was at it.

I bumped the version number in the pubspec. This is technically a
potentially breaking change because the signatures of createHitmap() and
mergeHitmaps() are more precise now. The only code I could find calling
that was in flutter_tools and that seems to be happy with the change.
If you'd rather I keep the old version, let me know.


cbracken:
Thank you, good sir!

PR #215, opened 2017-12-20
Enhance LCOV formatter.

DmitryKrutskikh:
Add number of instrumented lines 'LF' and number of lines with a non-zero execution count 'LH'

### DIFF
--- a/lib/src/formatter.dart
+++ b/lib/src/formatter.dart
@@ -50,11 +50,12 @@ class LcovFormatter implements Formatter {
       }
 
       buf.write('SF:$source\n');
-      v.keys.toList()
-        ..sort()
-        ..forEach((int k) {
-          buf.write('DA:$k,${v[k]}\n');
-        });
+      final lines = v.keys.toList()..sort();
+      lines.forEach((int k) {
+        buf.write('DA:$k,${v[k]}\n');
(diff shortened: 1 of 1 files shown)


DmitryKrutskikh:
@cbracken can you merge this PR?

lib/src/formatter.dart
```diff
@@ -50,11 +50,12 @@ class LcovFormatter implements Formatter {
...
       buf.write('SF:$source\n');
-      v.keys.toList()
-        ..sort()
-        ..forEach((int k) {
-          buf.write('DA:$k,${v[k]}\n');
-        });
+      final lines = v.keys.toList()..sort();
+      lines.forEach((int k) {
+        buf.write('DA:$k,${v[k]}\n');
+      });
```
cbracken: Use a for-loop here rather than `.forEach()`

https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal

cbracken: Actually never mind, I'll follow up and eliminate my previous shameful commit separately ;)

cbracken:
LGTM -- thanks for the patch and apologies for missing this earlier!

Issue #194, opened 2017-10-03
VMScript URI does not uniquely identify a script


cbracken:
In at least [one place](https://github.com/dart-lang/coverage/blob/0ca3888ca98cada168fdc47930aa5b7640904be2/lib/src/collect.dart#L85-L87) in the code, we use `VMScript.uri` as a proxy for the ID under the assumption that they map 1:1. 

This is not a safe assumption.

cbracken:
@rmacnak-google @jason-simmons

amirh:
This was failing the coverage tool when running on animated_icons_private_test, stack trace was:
```
unhandled error during test:
/b/build/slave/Linux/build/flutter/packages/flutter/test/material/animated_icons_private_test.dart
Bad state: Couldn't find line and column for token 246 in package:flutter/src/material/animated_icons/animated_icons.dart.
05:45 +860 ~2 -1: loading /b/build/slave/Linux/build/flutter/packages/flutter/test/material/animated_icons_private_test.dart [E]
  Bad state: Couldn't find line and column for token 246 in package:flutter/src/material/animated_icons/animated_icons.dart.
  package:flutter_tools/src/test/flutter_platform.dart 399   _FlutterPlatform._startTest
  ===== asynchronous gap ===========================
  dart:async/future_impl.dart 22                             _Completer.completeError
  package:flutter_tools/src/test/flutter_platform.dart 403   _FlutterPlatform._startTest
  ===== asynchronous gap ===========================
  dart:async/zone.dart 1045                                  _CustomZone.registerUnaryCallback
  dart:async-patch/async_patch.dart 30                       _asyncThenWrapperHelper
  package:flutter_tools/src/test/flutter_platform.dart 141   _FlutterPlatform._startTest
  package:flutter_tools/src/test/flutter_platform.dart 134   _FlutterPlatform.loadChannel
  package:test/src/runner/plugin/platform.dart 62            PlatformPlugin.load
  ===== asynchronous gap ===========================
  dart:async/zone.dart 1034                                  _CustomZone.registerCallback
  dart:async/zone.dart 924                                   _CustomZone.bindCallback
  dart:async/schedule_microtask.dart 148                     scheduleMicrotask
  dart:async/future.dart 198                                 new Future.microtask
  package:test/src/runner/plugin/platform.dart 59            PlatformPlugin.load
  package:test/src/runner/loader.dart 253                    Loader.loadFile.<fn>
  ===== asynchronous gap ===========================
[...]

PR #216, opened 2018-01-05
Use VMScriptRef to identify VM scripts, not URI

cbracken:
Previously, coverage collection assumed that a script could be uniquely
identified by URI, which is not a valid assumption. For example, when a
part is loaded via two libraries, both of which are loaded by an
isolate, the VM will track these are two scripts that map to the same
URI.

During collection, we now track each script by its (unique)
VMScriptRef. This ensures we lookup the correct script when computing
the affected line for each hit token. The hitmap remains URI based,
since in the end, we want a single, unified set of line->hitCount
mappings per script.

Fixes dart-lang/coverage#194


yjbanov:
lgtm

Issue #220, opened 2018-06-23
[SDK 2.0.0-dev.64.1] Unable to format the coverage as LCOV


cedx:
On Ubuntu 18.04 with Dart SDK 2.0.0-dev.64.1, the coverage collection is working fine, but when I try to format the JSON report as LCOV format, I always get this error:

```shell
/usr/lib/dart/bin/pub run coverage:format_coverage --in=var/coverage.json --lcov --out=var/lcov.info --packages=.packages --report-on=lib
  Unhandled exception:
  type 'List<dynamic>' is not a subtype of type 'List<Map<String, dynamic>>'
  #0      parseCoverage (package:coverage/src/hitmap.dart:76:66)
  <asynchronous suspension>
  #1      main (file:///home/cedric/.pub-cache/hosted/pub.dartlang.org/coverage-0.11.0/bin/format_coverage.dart:45:22)
  <asynchronous suspension>
  #2      _startIsolate.<anonymous closure> (dart:isolate/runtime/libisolate_patch.dart:277:32)
  #3      _RawReceivePortImpl._handleMessage (dart:isolate/runtime/libisolate_patch.dart:165:12)
```

cbracken:
This sounds like changes introduced in 2.0.0-dev.64.1; we ran into a lot of new type-check errors in Flutter with that roll as well.

Thanks for reporting! I'll take a look.

cbracken:
This is published as part of version 0.12.0, just released.

cedx:
Thanks @cbracken !

cbracken:
Anytime! Thanks again for reporting :) 

Probably not too relevant, but I've actually pushed out 0.12.1 as well which moves to Dart 2 constants, fixes a package:args deprecation warnings, and updates to a newer version, plus applies the latest dartfmt.

PR #221, opened 2018-06-25
Fix type mismatch error in parseCoverage

cbracken:
json.decode() returns a List<dynamic>, but the previous declaration
over-narrowed it to List<Map<String, dynamic>>. The coverage file format
is still expected to be a list of JSON maps, but we no longer declare
the local as narrowly.


cbracken:
Fixes #220.

PR #222, opened 2018-06-26
Update tests for Dart SDK 2.0.0-dev.64.1

cbracken:
The --checked flag has been removed in Dart SDK 2.0.0-dev.64.1. This
changes the default for the `checked` parameter in runAndCollect() to
false. In Dart 2, checked mode is no longer necessary, since type checks
are done statically at compile time. Support for the --checked flag will
be removed in a future version of package:coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased changes
+
+ * Strong mode fixes as of Dart SDK 2.0.0-dev.64.1.
+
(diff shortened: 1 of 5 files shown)


keertip:
Do we care about asserts here? If so you need to pass in the ` --enable-asserts` flag in dart 2 mode.

cbracken:
@keertip that's a good point. --checked also sets --enable-asserts; I can preserve that behaviour for the time being.

PR #236, opened 2018-10-13
Remove redundant asynchrony in api

DmitryKrutskikh:


kevmoo:
Nice!

lib/src/resolver.dart
```diff
@@ -158,9 +158,9 @@ class Loader {
 
   /// Loads an imported resource and returns a [Future] with a [List] of lines.
   /// Returns `null` if the resource could not be loaded.
-  Future<List<String>> load(String path) async {
+  List<String> load(String path) {
```
cbracken: The existing API which returns a `Future` of the lines in the source file is an intentional design decision to allow for an HTTP-based loader that loads files from the VM service port. Dart doesn't provide synchronous socket APIs, so an async API in required in such cases.

lib/src/resolver.dart
```diff
@@ -158,9 +158,9 @@ class Loader {
...
     try {
-      return new File(path).readAsLines();
+      return new File(path).readAsLinesSync();
```
cbracken: Since I/O operations are slow, and this is called in a loop over potentially thousands of source files, we want this to be async in order to allow the VM to continue processing lines while we wait for the underlying OS syscalls to load the next source(s).

lib/src/resolver.dart
cbracken: A potential improvement over the current API would be to switch to a `Stream<String>`, then read the file in as a byte-stream, and run it through a `UTF8Decoder` and a `LineSplitter`.

lib/src/resolver.dart
cbracken: By the way, if you have measured and *are* seeing performance improvements running on source sets with on the order of thousands of files, I'd be very curious to know!

PR #250, opened 2019-03-17
Collect coverage data on isolate exit

simolus3:
This implements the `on-exit` flag, which watches all isolates on the vm service and extracts coverage information when they exit. The collection process will complete when all isolates have finished.

This makes it easier to collect coverage for VM tests and builders. Instead of manually having to write test bundles that call all the `main` methods of test files sequentially in a single isolate, coverage can be collected on this script using `--on-exit`:
```dart
import 'package:test_core/src/executable.dart' as test;
import '../.dart_tool/build/entrypoint/build.dart' as builder;

void main() async {
  // Run the build runner
  await builder.main(['build', '--delete-conflicting-outputs']);
  // Run tests
  await test.main([]);
}
```
It wasn't possible to do this previously as the test runner will spawn another isolate for each test file, and the main isolate will only complete after the isolate for every test has completed. So, when running with `pause-isolates-on-exit` and `wait-paused`, the program will never reach a state in which all isolates are paused. However, if we collect coverage for each isolate as soon as it entered the `kPauseExit` state, the script will terminate after the coverage info for each isolate has been collected: We don't have to wait for _every_ isolate to be paused before starting to collect.

I extracted the collection process into two classes: `_OneTimeCollector`, which behaves like the existing process and `_OnExitCollector` which collects coverage whenever an isolate exits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ collected, it will wait until it detects a VM observatory to which it can
 connect. An optional `--connect-timeout` may be specified (in seconds).  The
 `--wait-paused` flag may be enabled, causing `collect_coverage` to wait until
 all isolates are paused before collecting coverage.
+Instead of waiting for all isolates to be paused, the `--on-exit` flag can be
+used to collect coverage whenever any isolate exits.
 
 #### Formatting coverage data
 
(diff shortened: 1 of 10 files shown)


cbracken:
Thanks for your PR and apologies for the delay in reviewing this. Due to other commitments, it'll be early next week before I get a chance to give this a proper review, but should be good to have a look then!

simolus3:
@kevmoo @cbracken I understand that you also have other priorities, but may I kindly ask you if there's any update on when this will get reviewed?
I think there is some interest in running coverage on the regular VM test runner and having the `--on-exit` flag would make that much easier.  If you know a simpler way to coverage the test runner please let me know, maybe we can also close this and I'll use that approach, thanks!

simolus3:
This is outdated given that the test package can now collect coverage out of the box.

PR #253, opened 2019-05-14
Remove optional 'new' keyword

cbracken:
The new keyword was made optional in Dart 2.0.0 and is no longer
necessary.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -71,7 +71,9 @@ linter:
     - type_annotate_public_apis # subset of always_specify_types
     - type_init_formals
     - unnecessary_brace_in_string_interps
+    - unnecessary_const
     - unnecessary_getters_setters
+    - unnecessary_new
 
     # === pub rules ===
     - package_names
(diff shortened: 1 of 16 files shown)


kevmoo:
Did you run `dartfmt --fix -w .` – for all of the things?
Enable the associated lints?

// --fix-doc-comments               Use triple slash for documentation comments.
'slash_for_doc_comments',
// --fix-named-default-separator    Use "=" as the separator for named parameter default values.
'prefer_equal_for_default_values',
// --fix-function-typedefs          Use new syntax for function type typedefs.
'prefer_generic_function_type_aliases',
// --fix-optional-const             Remove "const" keyword inside constant context.
'unnecessary_const',
// --fix-optional-new               Remove "new" keyword.
'unnecessary_new',

cbracken:
Interesting -- looks like dartfmt on my machine (2.3.0-dev.0.1) is giving different results from travis :(

cbracken:
Or... I'm running on a different branch :)

PR #257, opened 2019-07-09
Skip dart: sources by default

cbracken:
Skips collecting coverage for VM scripts with URIs with scheme 'dart'.
There's a lot of this source, most users aren't interested in collecting
coverage for it, and those who do, can still get at it by passing the
`--include-dart` flag if they want it.

This cuts test runtime down to 36% of what it previously was.


cbracken:
> (Googler, please make sure the reason for overriding the CLA status is clearly documented in these comments.)

Reason is that the PR is from me, and I'm a Googler. CLA bot appears to be confused.

PR #258, opened 2019-07-09
Disable websocket compression on service port

cbracken:
This disables websocket compression when connecting to the VM service
port. Since almost all invocations of coverage collection tools happen
over a loopback connection to localhost, websocket compression uses a
ton of CPU, which causes a performance hit rather than a performance
benefit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## dev
 
+ * Disables WebSocket compression for coverage collection. Since almost all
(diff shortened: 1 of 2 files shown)


cbracken:
> (Googler, please make sure the reason for overriding the CLA status is clearly documented in these comments.)

The author of this PR is me, and I'm a Google employee.

cbracken:
Fixes #258.

PR #260, opened 2019-07-18
Handle null coverage range

grouma:
I'm not sure why the VM would return empty coverage information for a given range but it does in practice. Handle this situation so we don't randomly choke.


grouma:
cc @jonahwilliams

cbracken:
LGTM stamp from a Japanese personal seal

Thanks!

PR #262, opened 2019-07-23
Support scoped output

grouma:
- Add new option `--scope-output` which restricts coverage output
  - This significantly improves performance when provided
- Add corresponding test
- Updated minimum sdk given the new service API call that only existed in `2.1.1+`

 cc @jonahwilliams

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:vm_service_lib/vm_service_lib.dart';
+
 import 'util.dart';
 
 const _retryInterval = Duration(milliseconds: 200);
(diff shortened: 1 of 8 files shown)


lib/src/collect.dart
```diff
@@ -65,19 +66,37 @@ Future<Map<String, dynamic>> collect(
...
-    allCoverage.addAll(coverage);
+    if (scopedOutput.isNotEmpty) {
+      final scripts = await service.getScripts(isolateRef.id);
+      for (var script in scripts.scripts) {
+        final uri = Uri.parse(script.uri);
+        if (uri.scheme != 'package') continue;
+        final scope = uri.path.split('/').first;
+        // Skip scripts which should not be included in the report.
+        if (!scopedOutput.contains(scope)) continue;
+        final scriptReport = await service.getSourceReport(
```
jonahwilliams: This may not matter too much, but you could try and run a few concurrently by building up a list of futures and then awaiting the result. Something like.

var pending = <Future<void>>[];
...
pending.add(service.getSourceReport(...).then((scriptReport) {
  return _getCoverageJson(....);
}).then(allCoverage.addAll));
...
await Future.wait(pending);

grouma: Doing some internal testing this suggestion surprisingly seemed to slow things down.

PR #270, opened 2019-09-26
Filter coverage by isolate ID

### LINKED ISSUE #269: Coverage collection is not supported for a single isolate
As of right now, coverage is gathered via all isolates, which happens here:
https://github.com/dart-lang/coverage/blob/master/lib/src/collect.dart#L81

I think it is valid to collect coverage from a single isolate (or set of isolates), which could be done by simply adding it as a named parameter in the `collect` function.

willdrach-wk:
Closes #269 

FYI: @grouma

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -33,7 +33,7 @@ const _retryInterval = Duration(milliseconds: 200);
 /// scripts that start with any of the provided paths are considered.
 Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
     bool waitPaused, bool includeDart, Set<String> scopedOutput,
-    {Duration timeout}) async {
+    {Duration timeout, Set<String> isolateIds}) async {
   scopedOutput ??= Set<String>();
   if (serviceUri == null) throw ArgumentError('serviceUri must not be null');
 
@@ -63,7 +63,7 @@ Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
       await _waitIsolatesPaused(service, timeout: timeout);
     }
 
-    return await _getAllCoverage(service, includeDart, scopedOutput);
+    return await _getAllCoverage(service, includeDart, scopedOutput, isolateIds);
   } finally {
     if (resume) {
       await _resumeIsolates(service);
@@ -73,12 +73,13 @@ Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
 }
 
 Future<Map<String, dynamic>> _getAllCoverage(
-    VmService service, bool includeDart, Set<String> scopedOutput) async {
+    VmService service, bool includeDart, Set<String> scopedOutput, Set<String> isolateIds) async {
   scopedOutput ??= Set<String>();
   final vm = await service.getVM();
   final allCoverage = <Map<String, dynamic>>[];
 
   for (var isolateRef in vm.isolates) {
+    if (!isolateIds.contains(isolateRef.id)) continue;
     if (scopedOutput.isNotEmpty) {
       final scripts = await service.getScripts(isolateRef.id);
       for (var script in scripts.scripts) {
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -31,9 +31,12 @@ const _retryInterval = Duration(milliseconds: 200);
 ///
 /// If [scopedOutput] is non-empty, coverage will be restricted so that only
 /// scripts that start with any of the provided paths are considered.
+///
+/// if [isolateIds] is set, the coverage gathering will be restricted to only
+/// those VM isolates.
 Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
     bool waitPaused, bool includeDart, Set<String> scopedOutput,
-    {Duration timeout, Set<String> isolateIds}) async {
+    {Set<String> isolateIds, Duration timeout}) async {
   scopedOutput ??= Set<String>();
   if (serviceUri == null) throw ArgumentError('serviceUri must not be null');
 
@@ -63,7 +66,8 @@ Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
       await _waitIsolatesPaused(service, timeout: timeout);
     }
 
-    return await _getAllCoverage(service, includeDart, scopedOutput, isolateIds);
+    return await _getAllCoverage(
+        service, includeDart, scopedOutput, isolateIds);
   } finally {
     if (resume) {
       await _resumeIsolates(service);
@@ -72,14 +76,14 @@ Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
   }
 }
 
-Future<Map<String, dynamic>> _getAllCoverage(
-    VmService service, bool includeDart, Set<String> scopedOutput, Set<String> isolateIds) async {
+Future<Map<String, dynamic>> _getAllCoverage(VmService service,
+    bool includeDart, Set<String> scopedOutput, Set<String> isolateIds) async {
   scopedOutput ??= Set<String>();
   final vm = await service.getVM();
   final allCoverage = <Map<String, dynamic>>[];
 
   for (var isolateRef in vm.isolates) {
-    if (!isolateIds.contains(isolateRef.id)) continue;
+    if (isolateIds != null && !isolateIds.contains(isolateRef.id)) continue;
     if (scopedOutput.isNotEmpty) {
       final scripts = await service.getScripts(isolateRef.id);
       for (var script in scripts.scripts) {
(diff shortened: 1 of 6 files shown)


grouma:
We'll want a `CHANGELOG` and `pubspec` update. Also doc comment entry for the new optional parameter here: https://github.com/dart-lang/coverage/blob/8e5d629529bb8ebc6bb610f5444467272679bf9d/lib/src/collect.dart#L33

Probably doesn't hurt to add a simple test as well.

willdrach-wk:
On it 👍

lib/src/collect.dart
```diff
@@ -33,7 +33,7 @@ const _retryInterval = Duration(milliseconds: 200);
 /// scripts that start with any of the provided paths are considered.
 Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
     bool waitPaused, bool includeDart, Set<String> scopedOutput,
-    {Duration timeout}) async {
+    {Duration timeout, Set<String> isolateIds}) async {
```
evanweible-wf: nit: alphabetize the named params

lib/src/collect.dart
```diff
@@ -73,12 +73,13 @@ Future<Map<String, dynamic>> collect(Uri serviceUri, bool resume,
 }
 
 Future<Map<String, dynamic>> _getAllCoverage(
-    VmService service, bool includeDart, Set<String> scopedOutput) async {
+    VmService service, bool includeDart, Set<String> scopedOutput, Set<String> isolateIds) async {
```
evanweible-wf: I think `isolateIds` should probably be an optional param here as well, and then when using it below, only skip an isolate if `isolateIds` is non-null and contains that isolate's ID. As written now I think you'd get a NPE if a value is not provided to `isolateIds` when calling `collect()`

willdrach-wk: On the first point: there's only 1 place that `_getAllCoverage` is called, so it making it optional would just be unnecessarily wordy.

On the second point: fixed!

lib/src/collect.dart
willdrach-wk: Fixed

willdrach-wk:
@grouma this should be closer to being ready, but it appears there's tests failing in master as well:
https://travis-ci.org/dart-lang/coverage/jobs/589704245

lib/src/collect.dart
```diff
@@ -31,9 +31,12 @@ const _retryInterval = Duration(milliseconds: 200);
 ///
 /// If [scopedOutput] is non-empty, coverage will be restricted so that only
 /// scripts that start with any of the provided paths are considered.
+///
+/// if [isolateIds] is set, the coverage gathering will be restricted to only
```
grouma: [nit] Capital `I`

test/collect_coverage_api_test.dart
```diff
@@ -92,6 +101,11 @@ Future<Map<String, dynamic>> _collectCoverage(
     }
   });
   final Uri serviceUri = await serviceUriCompleter.future;
+  Set isolateIdSet;
```
cbracken: Add the generic to the type annotation here.

willdrach-wk: Done 👍

test/collect_coverage_api_test.dart
```diff
@@ -92,6 +101,11 @@ Future<Map<String, dynamic>> _collectCoverage(
...
+  if (isolateIds) {
+    isolateIdSet = Set<String>.from(<String>[]);
+  }
```
cbracken: Merge this assignment the declaration  and make the variable final; also prefer the default constructor for the empty set. I'd have recommended the new set literal syntax but it's not available until Dart 2.2 and coverage is still lower-bounded by SDK 2.1.1.
```
final Set<String> isolateIdSet = isolateIds ? Set() : null;
``

Comment (willdrach-wk) [reply test/collect_coverage_api_test.dart]: Done

File: test/collect_coverage_api_test.dart [hunk 6]
```diff
@@ -92,6 +101,11 @@ Future<Map<String, dynamic>> _collectCoverage(
...
   });
   final Uri serviceUri = await serviceUriCompleter.future;
+  Set<String> isolateIdSet;
+  if (isolateIds) {
+    isolateIdSet = Set<String>.from(<String>[]);
+  }
 
-  return collect(serviceUri, true, true, false, scopedOutput, timeout: timeout);
+  return collect(serviceUri, true, true, false, scopedOutput,
+      timeout: timeout, isolateIds: isolateIdSet);
```
cbracken: This is a great start, but we should also add a test wherein we run with multiple isolates and only collect for one of them so that we don't accidentally break/regress this in future.

willdrach-wk: Ok, I can look into doing that sometime today! I'll ping you when I have something figured out.

willdrach-wk:
Thanks Chris! I have some extra capacity today so I should be able to get something to you this afternoon.

test/collect_coverage_api_test.dart
willdrach-wk: @cbracken it looks like the test app was already spawning an isolate. I added a test to test against just that isolate instead of testing against both that isolate and root, and everything looks good. Let me know if you want any further changes!

test/collect_coverage_api_test.dart
```diff
@@ -99,10 +115,27 @@ Future<Map<String, dynamic>> _collectCoverage(
...
+    if (line.contains('isolateId = ')) {
+      isolateIdCompleter.complete(line.split(' = ')[1]);
+    }
   });
+
   final Uri serviceUri = await serviceUriCompleter.future;
-  final Set<String> isolateIdSet = isolateIds ? Set() : null;
+  final String isolateId = await isolateIdCompleter.future;
+  final Set<String> isolateIdSet =
+      isolateIds ? Set.from(<String>[isolateId]) : null;
```
cbracken: Prefer the `Set<E>.of(Iterable<E>)` [constructor](https://api.dartlang.org/stable/2.5.1/dart-core/Set/Set.of.html) introduced in Dart 2, which has the nice benefit of type inference, so you can just do `Set.of([isolateId])` without explicitly having to specify the generic.

willdrach-wk: Good to know! Updated!

willdrach-wk:
Nit addressed! Any timeline on getting the tests in master fixed @cbracken?

cbracken:
Thanks!

@grouma @willdrach-wk re: the coverage failures, this looks like an implementation change in the latest dev VM. I'll take a quick look to see if the diff seems sane and if not, will have a look at `git blame` to see what's up. 

This happens occasionally with both improvements and regressions. No need to block submission on that test failure.

cbracken:
Got it. Looks like the order of isolates being reported has changed in the latest VM. Was going to comment earlier something along the lines of "maybe we should add a note stating that the order the isolates is returned is implementation-defined, so if the tests ever break, we'll need to fix this." but I think the right solution is for us to tag the isolates we care about using [Isolate.debugName](https://api.dartlang.org/stable/2.5.1/dart-isolate/Isolate/debugName.html), which can be set when we spawn the isolate.

cbracken:
@willdrach-wk if you're interested in taking that on, then awesome. If not, I'm more than happy to put together a quick PR. I've been working on a fair bit of non-code related work the last couple weeks and am antsy to get some patches landed, so more than happy to do it, if that's the case :)

willdrach-wk:
@cbracken I'm dropping off for the day soon, so if you have some bandwidth to tackle it that would be great!

cbracken:
Will do!

cbracken:
> Googler, please make sure the reason for overriding the CLA status is clearly documented in these comments.

Additional commits are minor refactorings applied by @cbracken (Googler) to get tests passing against the latest dev Dart SDK.

cbracken:
@grouma mind having a quick review of my two patches?

cbracken:
Also -- @evanweible-wf thanks for reviewing! It's awesome when patches come in pre-reviewed!

cbracken:
> Googler, please make sure the reason for overriding the CLA status is clearly documented in these comments.

Same as moments ago. @cbracken (Googler/me) fixing an error I'd intentionally added to test a change to an error message string :)

cbracken:
In terms of the one remaining test failure, it looks like a change in the way the VM is computing coverage. In the stable VM and in 2.6.0-dev.0.0, I'm seeing `test_app_isolate.dart` [line 21](https://github.com/dart-lang/coverage/blob/8e5d629529bb8ebc6bb610f5444467272679bf9d/test/test_files/test_app_isolate.dart#L21) as not-hittable. However, in 2.6.0-dev.3.0, it appears it's being [marked as hit](https://travis-ci.org/dart-lang/coverage/jobs/590591640).

I'm only seeing this for that one function. The only thing that stands out is that it's async. As an experiment, I tried making the function synchronous instead of async, and the issue went away.

Since this doesn't indicate a bug in the coverage package, I'm fine submitting this patch as-is. I'll file a Dart VM issue and cross-reference it here.

/cc @a-siva @ZichangG

cbracken:
Filed  https://github.com/dart-lang/sdk/issues/38628

PR #273, opened 2019-10-09
Bump version of vm_service

Stargator:
I noticed due to the recent update in `test` that coverage depends on a older version of vm_service where as `test` uses the latest.

One option, `coverage` could drop `vm_service` from pubspec.yaml and just depend on the transitive version used by `test`. If interest, I'll make that change to this PR.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: coverage
-version: 0.13.3
+version: 0.13.4
 author: Dart Team <misc@dartlang.org>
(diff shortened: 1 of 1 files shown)


Stargator:
@cbracken or @kevmoo, could you please add the label `invalid` to this PR?

It will show Digital Ocean that this was [not a valid PR for the Hacktoberest challenge](https://hacktoberfest.digitalocean.com/faq#general). Since I moved the work to another PR, I wouldn't want this to count.

PR #274, opened 2019-10-09
Bump version of vm_service

Stargator:
I noticed due to the recent update in test that coverage depends on a older version of vm_service where as test uses the latest.

One option, coverage could drop vm_service from pubspec.yaml and just depend on the transitive version used by test. If interest, I'll make that change to this PR.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -13,7 +13,7 @@ dependencies:
   package_config: '>=0.1.5 <2.0.0'
   path: '>=0.9.0 <2.0.0'
   stack_trace: ^1.3.0
-  vm_service: ^1.0.0
+  vm_service: '>=1.0.0 <3.0.0'
 
 dev_dependencies:
   test: ^1.0.0


cbracken:
This doesn't appear to actually change any of the version constraints.

Generally speaking it's fine for a given package to have looser version constraints on any given package than its dependencies -- in general, the looser the version constraint the easier it is to interop with any other given set of packages.

> One option, coverage could drop vm_service from pubspec.yaml

Packages should always declare any direct dependencies. Coverage needs to work with lots of packages independent independent of whether it's used with package:test or not.

Stargator:
@cbracken Sorry, after a rebase, I loss the intended change. I've resolved it.

I understand your point, which is why I left it in pubspec.yaml as I figured that left it open to error if test ever dropped use of `vm_service`.

cbracken:
Thanks. As noted above, it's not clear to me what problem tightening the version constraint is addressing. Since the coverage package interoperates with versions of vm_service >= 1.0.0, it already works just fine with v2.0.0.

If the `coverage` and `test` dependencies are both dependencies of your app, pub's constraint-solver will already narrow the set of allowable versions of `vm_service` to be the intersection of the ranges allowed by any dependant package. i.e. narrowing the constraint here to match that of package test is effectively a no-op for any case where you depend on both, so I can't see an upside.

This does however have the downside of potentially breaking existing users: imagine `package:foo` which depends (transitively) on `package:coverage` today and (transitively) on `vm_service` < 2.0.0, this would result in incompatibility and force authors of that package to patch their code.

The main valid reason to tighten the constraint would be if there are cases where package:coverage itself is actually broken when using vm_service versions >= 1.0.0 and < 2.0.0.

cbracken:
Sorry looking at this again I see the existing constraint is actually `^1.0.0` not `>=1.0.0`. I think the correct solution here is just to update this to `>=1.0.0 <3.0.0`.

pubspec.yaml
```diff
@@ -12,11 +12,11 @@ dependencies:
   logging: '>=0.9.0 <0.12.0'
   package_config: '>=0.1.5 <2.0.0'
   path: '>=0.9.0 <2.0.0'
-  stack_trace: ^1.3.0
-  vm_service: ^1.0.0
+  stack_trace: '^1.3.0'
```
kevmoo: Remove changes that just add gratuitous quotes.

Stargator: Sure, I intended it to keep it consistent with the rest of the file, but I'll remove the gratuitous quotes

Stargator:
@cbracken Since 2.0 of vm_service, there's a fix that would be helpful to have:
> Fixed issue where listening to EventStream.kHeapSnapshot and calling requestHeapSnapshot would throw an exception.

You're right in that it doesn't fix any errors breaking coverage, but by relying on an older version of a dependency the package and any users are missing out are fixes that could otherwise keep them protected from anyone trying to take advantage of known issues.

I'm just offering a suggested change, if it doesn't fit with the project's guidelines or isn't considered valuable to incorporate, okay. I'm not going to be hurt by it.

cbracken:
Thanks for the change. LGTM!

> You're right in that it doesn't fix any errors breaking coverage, but by relying on an older version of a dependency the package and any users are missing out are fixes that could otherwise keep them protected from anyone trying to take advantage of known issues.

Relying on the older version avoids breaking users who don't want to (or can't) upgrade to newer versions of the dep. This is common particularly in large, tightly-controlled codebases that have vetting processes that take days/weeks to vet open-source dependency updates. Users who need newer versions always have the option to apply tighter constraints to peg themselves to as precise a version range as they desire.

Issue #275, opened 2019-10-10
coverage considers const declarations as not visited, but they are assessed at compile time, not runtime


Justin-Randall:
When generating coverage data, const statements (say, for compile-time localized string constants or for many, many other cases where a const declaration is made), the code is considered as not covered. It can never be covered because it never really executes.

```dart
const String iLikeMyStringConstantsDRY = "Don't Repeat Yourself!";
```

This turns up as a miss under coverage. Declaring it final does not, but it then is something that is processed at run-time and is not truly, recursively const.

My use case: I have a CI build that runs unit tests that also generate coverage reports. New code should also have test coverage, except in cases where it is impossible to actually run (as in a const variable). 

To satisfy/pacify coverage reports, one would have to change the variable declaration to:
```dart
final String iLikeMyStringConstantsDRY = "Don't Repeat Yourself!";
```
but that would be incorrect and only add extra evaluation time to a program that does not need it.

cbracken:
Thanks for filing this; agreed this sounds like a bug.

/cc @a-siva for VM-side triage.

cbracken:
Spoke to @a-siva offline. The SDK side bug for this was https://github.com/dart-lang/sdk/issues/38284. This issue was fixed by @ZichangG in https://github.com/dart-lang/sdk/commit/3731cfdef9daafac6739569f3d5135f0cc66bedb. 

I'll add a test for this here so we can detect any regressions.

cbracken:
#277 adds a a const to one of our test apps in order to verify it doesn't appear in the coverage hitmap we produce.

PR #277, opened 2019-10-10
Test that const decls are not marked coverable

cbracken:
Constant declarations are evaluated during kernel compilation, not by
the VM. This patch adds a const to `test_app_isolate.dart` and verifies
that it doesn't appear in the coverage hitmap in
`collect_coverage_test.dart` and `run_and_collect_test.dart`.


cbracken:
Adds a test to catch regressions of #275.

Issue #278, opened 2019-11-30
vm_service 2.1.2 breaks coverage 0.13.3+1


mmcc007:
A recent upgrade of vm_service [breaks](https://travis-ci.com/mmcc007/screenshots/jobs/261763302#L328) coverage 0.13.3+1 (used to use [vm_service 2.1.1](https://travis-ci.com/mmcc007/screenshots/jobs/261154217#L342)). [Reverting](https://travis-ci.com/mmcc007/screenshots/jobs/261765381#L328) to coverage 0.13.3 resolves it for now ([vm_service 1.2.0](https://travis-ci.com/mmcc007/screenshots/jobs/261765381#L338)).

cbracken:
Discussed in person with @bkonyi. He's going to:
1. Publish vm_service 2.1.3 that has dart-lang/sdk@9e636b5 reverted. Since the current stable SDK doesn't yet include the IsolateGroup work, this'll fix anyone broken by that change.
2. Reland dart-lang/sdk@9e636b5 and publish a new vm_service release (2.2.0 maybe?) with a new SDK constraint that starts at whichever dev SDK first includes that support.

In the meantime, I'm going to quick publish a version of coverage with an upper bound of `<2.1.2` for `vm_service` to unbreak current users.

mmcc007:
Thanks for the quick response.

cbracken:
#279 has landed which sets the upper bound on the `vm_service` constraint to `<2.1.2` and has been published in version 0.13.3+2. 

Once items 1 and 2 have landed in the Dart SDK, I'll republish this package with an updated `vm_service` dependency.

cbracken:
Re-opening since there's still work to do :)

cbracken:
@mmcc007 thanks for reporting this! Please shout if you find the latest version on pub isn't working for you. @bkonyi mentioned it'll probably be a couple of days before he's able to get the upstream fixes landed in vm_service, but we'll try to get the long-term fix landed as quickly as possible.

mmcc007:
Will do. Thanks for the thorough support!

mmcc007:
FWIW: Perhaps including an integration test during the PR can avoid recurrences.

bkonyi:
A fix for package:vm_service is up for review here: https://dart-review.googlesource.com/c/sdk/+/127065

PR #279, opened 2019-12-03
Version 0.13.3+2

cbracken:
Tighten the dependency on the `vm_service` package from `>=1.0.0 <3.0.0`
down to `>=1.0.0 <2.1.2` in order to exclude version `2.1.2` which is
broken on the current stable Dart VM due to a missing SDK constraint in
its pubspec.yaml. The breakage was introduced in:
https://github.com/dart-lang/sdk/commit/9e636b5ab4de850fb19bc262e0686fdf14bfbfc0.


cbracken:
Fixes #278

cbracken:
CI passes on both stable and dev VMs.

PR #280, opened 2019-12-03
Loosen version restriction on package:vm_service now that crasher has been fixed

bkonyi:
See https://github.com/dart-lang/sdk/commit/7a911ce3f1e945f2cbd1967c6109127e3acbab5a

Fixes https://github.com/dart-lang/coverage/issues/278


cbracken:
In theory the constraint solver could still pick 2.1.2, but I'm not sure that pub provides a mechanism to exclude a version within a range.

PR #281, opened 2020-01-24
parseChromeCoverage

grouma:
Towards https://github.com/dart-lang/test/issues/36

Add `parseChromeCoverage` method to be used by `package:test` and internal tools. This method returns a hit-map based Dart report suitable for consumption by `package:coverage` to produce LCOV reports.


lib/src/chrome.dart
```diff
@@ -0,0 +1,151 @@
...
+  Future<String> Function(String scriptId) sourceProvider,
+  Future<String> Function(String scriptId) sourceMapProvider,
+  Future<Uri> Function(String sourceUrl, String scriptId) sourceUriProvider,
+) async {
+  final coverageReport = <Uri, Map<int, bool>>{};
+  for (Map<String, dynamic> entry in preciseCoverage) {
+    final String scriptId = entry['scriptId'];
+
+    final mapResponse = await sourceMapProvider(scriptId);
+    if (mapResponse == null) continue;
```
natebosch: [nit] Should we call out this behavior of ignoring values that don't get a source map or source file in the doc comment?

lib/src/chrome.dart
```diff
@@ -0,0 +1,151 @@
...
+    final offsetCoverage = _offsetCoverage(coverageInfo, compiledSource.length);
+    final coveredPositions = _coveredPositions(compiledSource, offsetCoverage);
+
+    for (var lineEntry in mapping.lines) {
+      for (var columnEntry in lineEntry.entries) {
+        if (columnEntry.sourceUrlId == null) continue;
+        final sourceUrl = mapping.urls[columnEntry.sourceUrlId];
+
+        // Ignore coverage information for the SDK.
+        if (sourceUrl.startsWith('org-dartlang-sdk:')) continue;
```
natebosch: How stable is this? Where is it defined?

lib/src/chrome.dart
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  final coverageHitMaps = <Uri, Map<int, int>>{};
+  coverageReport.forEach((uri, coverage) {
+    final hitMap = <int, int>{};
+    for (var line in coverage.keys.toList()..sort()) {
+      hitMap[line] = coverage[line] ? 1 : 0;
+    }
+    coverageHitMaps[uri] = hitMap;
+  });
```
natebosch: [optional]

```dart
final coverageHitMaps = {
  for(var report in coverageReport.entries)
    report.key: {
      for(var line in report.value.keys.toList()..sort())
        line: report.value[line] ? 1 : 0,
    },
};
```

lib/src/chrome.dart
```diff
@@ -0,0 +1,151 @@
...
+
+  final allCoverage = <Map<String, dynamic>>[];
+  coverageHitMaps.forEach((uri, hitMap) {
+    allCoverage.add(toScriptCoverageJson(uri, hitMap));
+  });
```
natebosch: [optional]

```dart
final allCoverage = [
  for(var hitMap in coverageHitMaps.entries)
    toScriptCoverageJson(hitMap.key, hitMap.value),
];
```

cbracken:
Exciting! Will take a look today!

lib/src/chrome.dart
```diff
@@ -0,0 +1,151 @@
...
+  final result = <_CoverageInfo>[];
+  for (Map<String, dynamic> functions in entry['functions']) {
+    for (Map<String, dynamic> range in functions['ranges']) {
+      result.add(_CoverageInfo(
+        range['startOffset'],
+        range['endOffset'],
+        range['count'] > 0,
+      ));
+    }
+  }
```
natebosch: [optional]

```dart
  return [
    for (Map<String, dynamic> functions in entry['functions'])
      for (Map<String, dynamic> range in functions['ranges'])
        _CoverageInfo(
          range['startOffset'],
          range['endOffset'],
          range['count'] > 0,
        ),
  ];
```

test/chrome_test.dart
```diff
@@ -0,0 +1,63 @@
...
+    final Map<int, int> expectedHits = {
+      5: 1,
+      6: 1,
+      7: 1,
+      8: 0,
+      10: 1,
+      11: 1,
+      13: 1,
+      14: 1,
+      15: 1,
```
natebosch: [nit] Would this be shorter without a trailing comma?

lib/src/chrome.dart
grouma: Done.

lib/src/chrome.dart
grouma: It comes from the SDK. I don't believe there is a way to programatically consume this value. We have similar logic in `package:dwds`.

lib/src/chrome.dart
grouma: The SDK constraint is wide on this package so I don't believe we can use this feature yet.

lib/src/chrome.dart
grouma: See above comment.

test/chrome_test.dart
grouma: No difference.

lib/src/chrome.dart
grouma: Going to leave this one as is.

lib/src/chrome.dart
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
```
cbracken: Not sure how this is done on the Dart team, but the general policy we use for Flutter is to use a single year for all files (in our case the earliest date we had in any of them, which was 2013). We used to use a policy that new files got the current year, old files preserved their headers, but that changed a few years back.

If current-year is still the right approach for Dart packages, then LGTM. Otherwise we should probably (in a separate PR), set these all consistently to 2013 in this repo, which is the earliest date I see in our headers.

lib/src/chrome.dart
```diff
@@ -0,0 +1,154 @@
...
+/// A covered position in a source file where [line] and [column] are 1 based.
+class _Position {
+  _Position(this.line, this.column);
+
+  final int line;
+  final int column;
+
+  @override
+  int get hashCode => hash2(line, column);
+  @override
```
cbracken: nit: add a  blank line before the `@override`.

lib/src/chrome.dart
grouma: Most other packages I touch follow the current-year pattern.

PR #283, opened 2020-01-25
Merge Coverage data

grouma:
I'm not sure if this situation will come up in practice but it is a good idea to be consistent with the behavior. If two source map locations point to the same Dart line, we will merge the coverage information.


cbracken:
Test?

grouma:
This is more precautionary. It's a little tricky to add a test. I'm not even sure this will show up in practice. Basically I need two entries in the source map that point to the same line and each need to differ in coverage.

@sigmundch is this possible? If so can you provide an example so that I can update the test fixtures.

sigmundch:
It may be possible with dart2js if the code is inlined in two places. Here is a sample program that uses some pragmas to control how dart2js does inlining:

```dart
int x = 0;

@pragma('dart2js:tryInline')
myInlinedMethod() {
  x++;
}

@pragma('dart2js:noInline')
int method1() {
  myInlinedMethod();
  return x;
}
@pragma('dart2js:noInline')
int method2() {
  myInlinedMethod();
  return x;
}

main() {
  int res = method1();
  if (res == 0) {
    method2();
  }
}
```

This will generate the following segment in the output JavaScript:
```js
Q = {
  method1: function() {
    return $.x = $.x + 1;
  },
  method2: function() {
    return $.x = $.x + 1;
  },
  main: function() {
    if (Q.method1() === 0)
      Q.method2();
  }
},
```

The body of method1 and method2 should have separate source-map entries pointing to the same code in myInlinedMethod. Because of the conditional, only the body in method1 is covered, but the one in method2 is not.

grouma:
I'm struggling to reproduce this scenario with an actual Dart2JS / DDC invocation. I'd still like to land the change though. PTAL.

grouma:
On second attempt I got it working with Siggi's suggestion. Updated the fixtures!

PR #286, opened 2020-02-04
3 commits: cleanup and fixes

kevmoo:


bin/format_coverage.dart
```diff
@@ -220,7 +220,8 @@ Environment parseArgs(List<String> arguments) {
 /// are contained by it if it is a directory, or a [List] containing the file if
 /// it is a file.
 List<File> filesToProcess(String absPath) {
-  final filePattern = RegExp(r'^dart-cov-\d+-\d+.json$');
+  print(absPath);
```
grouma: Remove?

lib/src/resolver.dart
```diff
@@ -157,7 +157,9 @@ class Loader {
   /// Returns `null` if the resource could not be loaded.
   Future<List<String>> load(String path) async {
     try {
-      return File(path).readAsLines();
+      final lines = await File(path).readAsLines();
+      // ensure we await `readAsLines` within the try block so errors are caught
```
grouma: [nit] Capitalize and end in a period.

bin/format_coverage.dart
```diff
@@ -220,7 +220,8 @@ Environment parseArgs(List<String> arguments) {
...
+  final filePattern = RegExp(r'^.*\.json$');
```
grouma: Add a simple test for this changed behavior?

lib/src/resolver.dart
kevmoo: ack

bin/format_coverage.dart
kevmoo: ack

bin/format_coverage.dart
kevmoo: going to remove this commit and send another PR w/ something better here

lib/src/resolver.dart
```diff
@@ -157,7 +157,9 @@ class Loader {
   /// Returns `null` if the resource could not be loaded.
   Future<List<String>> load(String path) async {
     try {
-      return File(path).readAsLines();
+      final lines = await File(path).readAsLines();
+      // Ensure `readAsLines` runs within the try block so errors are caught.
```
natebosch: I don't understand this comment

`return await File(path).readAsLines();` should be sufficient and clear. In a future SDK release the `await` won't be necessary.

test/chrome_test.dart
```diff
@@ -28,18 +28,17 @@ Future<Uri> sourceUriProvider(String sourceUrl, String scriptId) async =>
 
 void main() {
   test('reports correctly', () async {
-    final preciseCoverage = json.decode(
+    final List preciseCoverage = json.decode(
```
natebosch: This is not how we want to handle casts.

```dart
final preciseCoverage = json.decode(...) as List;
```

I'd prefer to fix the package analysis options than to expand use of implicit casts

lib/src/resolver.dart
kevmoo: ack!

test/chrome_test.dart
kevmoo: Agreed. Let's do that in a follow-up. There are a lot of things to clean up.

PR #287, opened 2020-02-05
Enable and fix pedantic lints

kevmoo:


test/run_and_collect_test.dart
```diff
@@ -43,8 +43,8 @@ void main() {
     final hitMap = createHitmap(coverage);
     expect(hitMap, contains(_sampleAppFileUri));
 
-    final Map<int, int> isolateFile = hitMap[_isolateLibFileUri];
-    final Map<int, int> expectedHits = {
+    final isolateFile = hitMap[_isolateLibFileUri];
```
cbracken: This change seems like a significant net negative in terms of readability.

cbracken: Not claiming that passing around these blobs of `Map<int, int>` is particularly readable in the first place ;-)

kevmoo: We're trying to standardize on pedantic lints. Many of us feel, with strong inference, explicit types are more noise then signal – it's tough to decide when to put them in.

cbracken: As noted in the comment above, I generally agree -- in most cases that is the case, and empirically the fact that lots of languages are adding support for `var`/`val`/`auto`/etc. suggests most of us agree on that. I'm pointing out that in this case it's a net negative with regard to readability. Not sure that there's much we can do about this without loading the code with `// ignore`s or turning off pedantic lints and trusting folks to write readable code.

For this case, one way to solve this without adding the types back would be the addition of new (non-map) types for hit-maps and the line-hitcount mappings, which avoids raw index operator lookups in place of a method that makes clear what's being fetched.

natebosch: In my opinion this version of the code is _more_ readable. If we state the goal of reading the code as "knowing what type `isolateFile` is", I agree this is less readable. I don't think that's an important detail while reading. If we state the goal of reading the code as "knowing what the intention  and conditions of the test are", I think this is either equally or more readable than the old code. Some variable renames would help even more with that.

natebosch: Does https://github.com/dart-lang/coverage/pull/288 address your concerns?

I agree that `isolateFile` wasn't a good name, I don't think the explicit type mitigated that much and is mostly orthogonal changes.

cbracken: Does indeed! Domo arigato!

Issue #272, opened 2019-10-01
Coverage json file pattern used when a directory is provided as the input to the formatter


evanweible-wf:
The `coverage:format_coverage` requires an `--in` option to find the coverage json input(s) to be formatted. As currently implemented, the value can be either:

1. A path to a single file
2. A path to a directory within which one or more coverage json files exist

@willdrach-wk is currently working on integrating coverage into the test runner, and in that scenario, a coverage output is created for each test file (and probably one per-environment, e.g. VM or Chrome). In this scenario, using #2 is obviously much more useful because we could point the formatter towards a directory with all of the coverage outputs and receive a single lcov artifact.

> Side note: it was recommended to us that we don't try to combine/merge coverage collected on different platforms, like VM and Chrome, because the lines/hitmaps might not line up perfectly. However, at least for VM coverage collection, merging multiple inputs works fine. TBD on Chrome coverage collection, but it seems like we could process the VM inputs and the Chrome inputs separately.

The way #2 is currently implemented, it will only look for files in the given directory that match this pattern:

```
final filePattern = RegExp(r'^dart-cov-\d+-\d+.json$');
```

Is there any particular reason this filename pattern was chosen? I ask because it unfortunately limits the readability of the files that the coverage collection step outputs. Will's current branch of the test pkg that collects VM coverage outputs one file per test and adds a `.vm.json` extension to the filename, which is nice because it's very clear which coverage artifact corresponds to which test, but it doesn't get picked up by this pattern.

I'm not sure if relaxing this pattern would even be acceptable or if that would be viewed as a breaking change – just thought I'd ask.

@cbracken

willdrach-wk:
For more context: the PR in question is dart-lang/test#1088

If I run the `test` package in the following way:
```bash
pub run test --coverage hello_world test/vm/simple_repo_test.dart
```

A coverage `.json` file gets outputted to `hello_world/test/vm/simple_repo_test.dart.vm.json`, and if there's multiple test suites/files, we get multiple coverage files out of it. It would be awesome to just point to that `hello_world` directory like Evan suggested.

evanweible-wf:
Any thoughts on this @cbracken?

cbracken:
Generally, I'd love for our tools to get out of the business of dealing with the intermediate JSON altogether (it's mostly just historical baggage and should be an implementation detail) -- we should just have APIs that people could implement to aggregate/filter what they care about. 

[...]

PR #290, opened 2020-02-10
Consider all json for format

grouma:
Closes https://github.com/dart-lang/coverage/issues/272

### DIFF
--- a/test/format_coverage_test.dart
+++ b/test/format_coverage_test.dart
@@ -1,0 +1,33 @@
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+import '../bin/format_coverage.dart';
+
+void main() {
+  Directory testDir;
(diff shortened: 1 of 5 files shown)


test/format_coverage_test.dart
```diff
@@ -0,0 +1,33 @@
...
+
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+import '../bin/format_coverage.dart';
+
+void main() {
+  Directory testDir;
+  setUp(() {
+    testDir = Directory.systemTemp.createTempSync('coverage_test_temp');
```
kevmoo: https://pub.dev/packages/test_descriptor does this dance really well – not sure if want to deal with that, though...

grouma: Yeah I thought about that but given that it will only be used for a single test I decided against adding the dep.

PR #292, opened 2020-02-14
update the Resolver api to use the new package config api

### LINKED ISSUE #459: Migrate to the latest package_config package
~This is needed to unblock package:test users~ Note that this is no longer blocking as we are doing a non-breaking release of package_config. But we should still update once it is available.

jakemac53:
Closes https://github.com/dart-lang/tools/issues/459.

Blocked on https://github.com/dart-lang/package_config/issues/68 and a published version of package_config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.13.7 - 2020-02-14
+
+* Update to package_config `1.9.0` which supports package_config.json
+  files and should be forwards compatible with `2.0.0`.
+
 ## 0.13.6 - 2020-02-10
 
 * Now consider all `.json` files for the `format_coverage` command.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Update to package_config `1.9.0` which supports package_config.json
   files and should be forwards compatible with `2.0.0`.
+* Deprecate the `packageRoot` argument on `Resolver`.
 
 ## 0.13.6 - 2020-02-10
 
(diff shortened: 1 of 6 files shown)


pubspec.yaml
```diff
@@ -1,15 +1,15 @@
 name: coverage
-version: 0.13.6
+version: 0.13.7
```
grouma: [optional] Should we just bump this to 1.0.0?

grouma (approved):

pubspec.yaml
jakemac53: this isn't breaking - I would do it separately afterwords or wait for NNBD?

grouma: Ah true. Let's just wait until NNBD.

natebosch (approved):

jakemac53:
PTAL, updated to rely on the new published version and fixed .packages parsing

cbracken (approved):
We should probably add a test here.

jakemac53:
Yep I am working on adding tests for a resolver with the new format - which also brought up an unexpected issue where the new package_config requires absolute file uris :D. Hooray for coverage :fireworks: .

cbracken:
You have all the LGTMs I have to give. Or at least all that matter per CL :)

jakemac53:
Going to do a final check through an internal global before landing

jakemac53:
Of course I have to land package_config internally first so this will be delayed likely until monday

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.13.7 - 2020-02-14
+
+* Update to package_config `1.9.0` which supports package_config.json
```
kevmoo: Note that `packageRoot` is now deprecated in the API?

jakemac53: done

PR #293, opened 2020-02-28
Make resumeIsolate robust, resilient in the face of exiting application.

aam:
When last isolate paused at exit resumes, the application might exit, close the socket.
If that happens coverage tool should handle closed socket and complete resume requests.


cbracken:
Looks like `dartfmt` is upset.

aam:
Thank you @cbracken .
On further testing on b/150193047 I tweaked it a little. PTAL.

PR #295, opened 2020-03-05
Use a more standard Dart + travis config

kevmoo:


natebosch:
Interesting, so you if you specify an `install` key, you lose the defaults you get from `language: dart`? That'll be good to remember.

kevmoo:
> Interesting, so you if you specify an `install` key, you lose the defaults you get from `language: dart`? That'll be good to remember.

yup!

PR #302, opened 2020-06-02
Add the ability to ignore lines from coverage depending on the comment

pasindud:
Use // coverage:ignore-line to ignore one line.
Use // coverage:ignore-start and // coverage:ignore-end to ignore range of lines inclusive.
Use // coverage:ignore-file to ignore the whole file.

### DIFF
--- a/bin/format_coverage.dart
+++ b/bin/format_coverage.dart
@@ -24,6 +24,7 @@ class Environment {
   bool prettyPrint;
   bool lcov;
   bool expectMarkers;
+  bool checkIgnore;
   bool verbose;
 }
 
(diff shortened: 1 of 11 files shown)


bin/format_coverage.dart
```diff
@@ -39,10 +40,16 @@ Future<Null> main(List<String> arguments) async {
     print('  package-root: ${env.pkgRoot}');
     print('  package-spec: ${env.packagesPath}');
     print('  report-on: ${env.reportOn}');
+    print('  check-ignore: ${env.checkIgnore}');
   }
 
   final clock = Stopwatch()..start();
-  final hitmap = await parseCoverage(files, env.workers);
+  print(env.checkIgnore);
```
grouma, line 47: Should this be removed?

pasindud:
Should we add a entry to README also ?

grouma:
> Should we add a entry to README also about this ?

Probably a good idea. Otherwise LGTM.

Elias8:
Is this available on the current stable which is `1.22.5`?

grouma:
This should be available in stable as it depends on version 14.1: https://github.com/flutter/flutter/blob/78910062997c3a836feee883712c241a5fd22983/packages/flutter_tools/pubspec.yaml#L17

lukepighetti:
I'm on Flutter `Channel dev, 1.26.0-8.0.pre` and have `// coverage:ignore-file` at the top of a file and it's still showing up in code coverage reports. Am I missing a configuration step?

grouma:
@lukepighetti I tried to reproduce your issue and I'm unable to do so. I simply ran `flutter test --coverage` on the `hello_world` example within the Flutter repo. I can add and ignore files as expected.

PR #304, opened 2020-06-04
Revert breaking change in 13.10

grouma:


CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
+## 0.13.11 - 2020-03-09
+
+* Revert breaking change in 13.10
+
 ## 0.13.10 - 2020-06-03
+
```
jakemac53, line 6: I would add a note that this is a broken release that was reverted here

PR #306, opened 2020-08-20
Remove unused dart:async imports

MichaelRFairhurst:
Since this version does not support Dart 1.0, these imports are
unnecessary.

### DIFF
--- a/lib/src/formatter.dart
+++ b/lib/src/formatter.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:async';
-
 import 'package:path/path.dart' as p;
 
 import 'resolver.dart';
--- a/lib/src/hitmap.dart
+++ b/lib/src/hitmap.dart
@@ -2,7 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:async';
 import 'dart:convert' show json;
 import 'dart:io';
 
--- a/lib/src/resolver.dart
+++ b/lib/src/resolver.dart
@@ -2,7 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:async';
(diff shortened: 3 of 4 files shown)


cbracken:
Presubmit appears to consistently fail.

jtmcdole:
`Future` is very clearly used in these files, so you have to have dart:async

cbracken:
IIRC Future and Stream were moved to `dart:core` and re-exported from `dart:async` sometime around Dart 2.0 or 2.1.

natebosch:
These timeouts happen locally for me on master too. @grouma @cbracken - do you know the intention of the specific tests? Have we had problems with them timing out before?

cbracken:
The three tests in question test:
* That when run format coverage into {LCOV,PrettyPrint} format the output includes coverage for files from the `reportOn` path list.
* That the PrettyPrint formatter emits coverage that looks right.

I've never seen those particular tests fail/time out in the past.

grouma:
> Have we had problems with them timing out before?

We have been broken by SDK changes in the past.  These particular tests have been failing for over 2 months.

The last passing SDK was:
`Dart SDK version: 2.9.0-21.0.dev (dev) (Wed Jul 8 10:54:08 2020 +0200) on "linux_x64"`

The first failing SDK as far as I can tell was:
`Dart SDK version: 2.10.0-0.0.dev (dev) (Wed Jul 22 19:58:38 2020 +0200) on "linux_x64"`

This doesn't look good and given how soon the SDK is about to drop we should investigate this issue with high priority.

@kevmoo do we not have alerts for this package?

jtmcdole:
@cbracken - 

https://github.com/dart-lang/sdk/blob/master/sdk/lib/async/future.dart#L148

natebosch:
@jtmcdole - https://github.com/dart-lang/sdk/blob/433fcd09d91677ec9a747506e2095f7c708a0672/sdk/lib/core/core.dart#L171-L172

jtmcdole:
I was thinking "exported" as in moved, not **actually** `export`ed :)

cbracken:
Was the underlying issue with DDC ever fixed before merging? I notice tests are [failing at head](https://travis-ci.org/github/dart-lang/coverage/jobs/728412753) but haven't looked at why.

Issue #309, opened 2020-09-14
Coverage tests are failing on the dev SDK at head


grouma:
See https://github.com/dart-lang/coverage/pull/306#issuecomment-692334586 for more context

grouma:
cc @natebosch @kevmoo

PR #311, opened 2020-09-21
Fix --wait-paused with no isolates

grouma:
Closes https://github.com/dart-lang/coverage/issues/309

There was a race condition in our logic which was exacerbated by SDK version `2.10`. If no isolates had started, we wouldn't properly wait to collect coverage information.


bkonyi (approved):

cbracken (commented):
LGTM!

PR #324, opened 2020-12-16
Remove packageRoot in run_and_collect

lukaslihotzki:
`--package-root=` is discontinued in dart.
(https://github.com/dart-lang/sdk/issues/41197)

PR #326, opened 2021-01-13
Version 0.15.0

cbracken:
Eliminates the `--package-root` option from both run_and_collect and
format_coverage commands, as well as from non-public API under lib/src.

Related patch #324 by @lukaslihotzki.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.15.0 - 2021-01-13
+
+* BREAKING CHANGE: Eliminate the `--package-root` option from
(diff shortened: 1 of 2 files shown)


cbracken:
Yuck. Looks like a circular dependency between test and coverage.

grouma:
This change doesn't seem to be impactful for `package:test` so we can preemptively allow for the next version. @natebosch thoughts?

grouma:
Closing favor of https://github.com/dart-lang/coverage/pull/327.

The preferred approach to deal with this type of cyclic dependency is to use a dependency override, remove before publishing and then update the cyclic dependency. This reduces the chances of issues and users won't get the latest version until both packages are updated.

PR #327, opened 2021-01-14
Prep to publish 0.15.0

grouma:
Eliminates the `--package-root` option from both run_and_collect and
format_coverage commands, as well as from non-public API under lib/src.

Related patch #324 by @lukaslihotzki.


natebosch (approved):

cbracken (approved):
lgtm

Issue #330, opened 2021-02-17
Ignoring via comments does not work at all


felix-barz-brickmakers:
Trying to exclude parts of a library from coverage with any of the comments will not work at all. Download the attached dart project [coverage_test.zip](https://github.com/dart-lang/coverage/files/5994405/coverage_test.zip) and run:
```.bash
$ dart test --coverage=coverage
$ dart run coverage:format_coverage --lcov --check-ignore \
		--in coverage \
		--out coverage/lcov.info \
		--packages .packages \
		--report-on lib
$ genhtml -o coverage/html coverage/lcov.info
```

The generated HTML will still contain all four test files will all lines. Nothing is ignored, even though all 3 variants have been tried out and `--check-ignore` was added to the command line.

iapicca:
@pasindud any update on the [issue](https://github.com/flutter/flutter/pull/61408#issuecomment-757839556)

> Thanks @felix-barz-brickmakers, I'll have to take a look at this issue.

pasindud:
So I think for non flutter stuff this commit should https://github.com/pasindud/coverage/commit/50fded8e59fa40713985ca166f404b92ca124a9b I'll have to test it bit more though


In the above example

```dart
class Foo<T> extends FooBase<T> {
  const Foo(T data)
      // coverage:ignore-line
      : super(data);

  T call() => data;
}
```


It should be 

```dart
class Foo<T> extends FooBase<T> {
  const Foo(T data)
      : super(data);        // coverage:ignore-line

  T call() => data;
}
```

mark-revision:
After upgrading to Flutter 2.0/Dart 2.12.0 this feature works perfectly and correctly ignores untestable lines.

grouma:
@iapicca what version of Flutter are you using? As @mark-revision stated, this should be working on the latest stable.

iapicca:
@grouma 
the issue seems to persist running the same `code` and `test` provided [above](https://github.com/dart-lang/coverage/issues/330#issuecomment-795507589)

<details>
<summary>pubspec</summary>

```yaml
name: issue_330
description: A new Flutter project.
version: 1.0.0+1

environment:
  sdk: '>=2.12.0 <3.0.0'

#dependencies:
#  flutter:
#    sdk: flutter


dev_dependencies:
#  flutter_test:
#    sdk: flutter
  test: ^1.16.7
```

</details>

<details>
<summary>logs</summary>

```bash
[...]

PR #336, opened 2021-03-16
Pass packages argument to format_coverage

grouma:
Closes https://github.com/dart-lang/coverage/issues/330

### DIFF
--- a/test/run_and_collect_test.dart
+++ b/test/run_and_collect_test.dart
@@ -2,8 +2,6 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:io';
-
 import 'package:coverage/coverage.dart';
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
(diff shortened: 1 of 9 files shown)


cbracken:
[collect_coverage_test.dar](https://github.com/dart-lang/coverage/blob/master/test/collect_coverage_test.dart) is an integration test, you could probably adapt some of the code in there to capture coverage then feed it to the formatter. Specifically see `_collectCoverage()` here https://github.com/dart-lang/coverage/blob/92ab6ea5d62ef913c60f9d91ecb55766c631cb18/test/collect_coverage_test.dart#L119

test/run_and_collect_test.dart
```diff
@@ -40,42 +38,30 @@ void main() {
...
-    if (Platform.version.startsWith('1.')) {
-      expectedHits[23] = 0;
-    } else {
-      // Dart VMs version 2.0.0-dev.6.0 mark the opening brace of a function as
-      // coverable.
-      expectedHits[11] = 1;
-      expectedHits[18] = 1;
-      expectedHits[28] = 1;
-      expectedHits[32] = 3;
-    }
```
cbracken, line 78: Haha fair enough; it's about time. We should probably add a note to the changelog about this removal of support.

cbracken (approved):
lgtm -- didn't hand count the expected hit counts, so I'm taking your work for it ;-)

LGTM stamp from a Japanese personal seal

PR #346, opened 2021-05-25
Support vm_service 7.0.0

bkonyi:


jakemac53 (approved):
LGTM, you can go ahead and remove the override now and we can land it.

PR #347, opened 2021-06-08
Return a sorted hitmap

grouma:
Update `createHitmap` to return a sorted hitmap which is required by the ignore line annotations.

### DIFF
--- a/lib/src/hitmap.dart
+++ b/lib/src/hitmap.dart
@@ -8,6 +8,21 @@ import 'dart:io';
 import 'package:coverage/src/resolver.dart';
 import 'package:coverage/src/util.dart';
 
+/// Class containing information about a coverage hit.
+class _HitInfo {
+  _HitInfo(this.firstLine, this.hitRange, this.hitCount);
+
+  /// The line number of the first line of this hit range.
+  final int firstLine;
+
+  /// A hit range is either a number (1 line) or a String of the form
+  /// "start-end" (multi-line range).
+  final dynamic hitRange;
+
+  /// How many times this hit range was executed.
+  final int hitCount;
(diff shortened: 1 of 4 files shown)


lib/src/hitmap.dart
```diff
@@ -80,7 +95,9 @@ Future<Map<String, Map<int, int>>> createHitmap(
     }
 
     final sourceHitMap = globalHitMap.putIfAbsent(source, () => <int, int>{});
-    final hits = e['hits'] as List;
+    var hits = e['hits'] as List;
+    // Ignore line annotations require hits to be sorted.
+    hits = _sortHits(hits);
```
liamappelbe, line 100: Since you're going to the trouble of parsing `hits` into `structuredHits`, which is a much nicer format, I think it would make more sense to leave it in that format for the remainder of this function, rather than transforming it back into the weird flat array.

So in the `_HitInfo` class you could remove `hitRange` and add a `int lastLine` (and set `lastLine` to be the same as `firstLine` in the case that `hitRange` is an `int`).

liamappelbe (changes requested):

lib/src/hitmap.dart
thammaknot: Would it be okay to do that in a separate PR?
I feel that making that change here would be a bit too much.
Changing the underlying data structure was definitely not the main objective of the change.

liamappelbe: I think you'll only have to delete a few lines of code from the loop below, and tweak _sortHits a bit. But I suppose it's fine to do that in a follow up.

liamappelbe (approved):

PR #348, opened 2021-06-11
Function coverage

liamappelbe:


lib/src/collect.dart
```diff
@@ -227,26 +270,47 @@ Future<List<Map<String, dynamic>>> _getCoverageJson(VmService service,
...
   final coverage = <Map<String, dynamic>>[];
-  hitMaps.forEach((uri, hitMap) {
-    coverage.add(toScriptCoverageJson(uri, hitMap));
+  hitMaps.forEach((uri, hits) {
+    coverage.add(toScriptCoverageJson(uri, hits));
   });
   return coverage;
 }
 
+void _increment(Map<int, int> counter, int key) {
```
grouma, line 297: Can you make this more descriptive? I had to look up what it was doing.

lib/src/hitmap.dart
```diff
@@ -8,6 +8,18 @@ import 'dart:io';
 import 'package:coverage/src/resolver.dart';
 import 'package:coverage/src/util.dart';
 
+/// Contains line and function hit information for a single script.
+class HitMap {
+  /// Map from line to hit count for that line.
+  final lineHits = <int, int>{};
+
+  /// Map from function definition line to hit count for that function.
```
grouma, line 16: What does a `function definition line` mean? Is that an ID?

lib/src/hitmap.dart
```diff
@@ -173,6 +205,32 @@ Future<Map<String, Map<int, int>>> parseCoverage(
   return globalHitmap;
 }
 
+/// Returns a JSON hit map backward-compatible with pre-1.16.0 SDKs.
```
grouma, line 208: Should we even support that low of an SDK? I'd rather just remove this logic and bump the min SDK.

lib/src/collect.dart
liamappelbe: Done

lib/src/hitmap.dart
liamappelbe: I'm fine with bumping the min SDK, but I don't know what logic in here is specific to older SDKs, so I'm not sure what changes you're asking for. Maybe we could do that in a follow up CL.

liamappelbe:
I think it might make sense to put function coverage behind a runtime flag. Gathering all the function info might slow things down, and for some users coverage is already irritatingly slow.

lib/src/hitmap.dart
grouma: I'm not sure either. That's a good question for @cbracken

grouma:
> I think it might make sense to put function coverage behind a runtime flag. Gathering all the function info might slow things down, and for some users coverage is already irritatingly slow.

Yeah that makes sense. I'd imagine it would be turned off by default for internal customers.

grouma (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+* Support function level coverage information, when running tests in the Dart
+   VM. This is not supported for web tests yet.
+* Add flag `--function-coverage` (abbr `-f`) to collect_coverage that collects
+  function coverage information.
+* Add flag `--pretty-print-func` (abbr `-f`) to format_coverage that works
+  similarly to pretty print, but outputs function level coverage, rather than
+  line level.
+* Update `--lcov` (abbr `-l`) in format_coverage to output function level
+  coverage, in addition to line level.
+* BREAKING CHANGE: The signatures of `createHitmap`, `mergeHitmaps`,
```
natebosch, line 12: breaking changes need major version bumps. This should be going out as `2.0.0`.

https://github.com/dart-lang/coverage/pull/355

PR #349, opened 2021-08-30
Remove references to travis

natebosch:
Remove `travis_report_failure.sh,` it is unreferenced and if we enable
failure reports on github actions we won't use a separate script.

Rename `travis.sh` to `test_and_collect.sh` to describe the behavior
instead of the use case.

Fixup the changelog. `1.0.4` was not published and will not be
published, so merge the changelog in with `1.1.0`. Append a `-dev` while
we are in between published states.


kevmoo (approved):

kevmoo:
Good catch on the changelog!

PR #350, opened 2021-09-01
Use the report lines flag

liamappelbe:
The report lines flag returns the line numbers directly in the source report, rather than the token positions. This lets us skip the extra RPC to get the token pos to line number mapping. In my benchmarks, this halved the time it takes to collect coverage.

We still have to get the script objects if function coverage was requested. So to make sure the RPC skipping flow is tested, I've updated some of the tests so that function coverage is tested separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,3 @@
-## 1.1.1-dev
-
-* Use the `reportLines` flag in `vm_service`'s `getSourceReport` RPC. This
-  typiclly halves the number of RPCs that the coverage collector needs to run.
-
 ## 1.1.0-dev
 
 * Support function level coverage information, when running tests in the Dart
@@ -21,6 +16,8 @@
   optional bool flag controlling whether function coverage is collected.
 * Ensure `createHitmap` returns a sorted hitmap. This fixes a potential issue with
   ignore line annotations.
+* Use the `reportLines` flag in `vm_service`'s `getSourceReport` RPC. This
+  typiclly halves the number of RPCs that the coverage collector needs to run.
 
 ## 1.0.3 - 2021-05-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 * Ensure `createHitmap` returns a sorted hitmap. This fixes a potential issue with
   ignore line annotations.
 * Use the `reportLines` flag in `vm_service`'s `getSourceReport` RPC. This
-  typiclly halves the number of RPCs that the coverage collector needs to run.
+  typically halves the number of RPCs that the coverage collector needs to run.
 
 ## 1.0.3 - 2021-05-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.1.1-dev
+
+* Use the `reportLines` flag in `vm_service`'s `getSourceReport` RPC. This
+  typiclly halves the number of RPCs that the coverage collector needs to run.
+
 ## 1.1.0-dev
 
 * Support function level coverage information, when running tests in the Dart
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -97,8 +97,10 @@ Future<Map<String, dynamic>> _getAllCoverage(
   scopedOutput ??= <String>{};
   final vm = await service.getVM();
   final allCoverage = <Map<String, dynamic>>[];
-  final ver = await service.getVersion();
-  final reportLines = ver.major == 3 && ver.minor != null && ver.minor! >= 51;
+  final version = await service.getVersion();
+  final reportLines =
+      (version.major == 3 && version.minor != null && version.minor! >= 51) ||
+          (version.major != null && version.major! > 3);
 
   for (var isolateRef in vm.isolates!) {
     if (isolateIds != null && !isolateIds.contains(isolateRef.id)) continue;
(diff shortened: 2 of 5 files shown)


kevmoo:
Seems like we can ignore the coveralls failure here...

grouma (commented):
We should have a changelog entry and update the user facing documentation if we haven't already.

liamappelbe:
> We should have a changelog entry and update the user facing documentation if we haven't already.

I've updated the changelog, but I'm not user what user facing documentation needs to be updated. If everything works correctly, it should be transparent to the user.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.1.1-dev
```
jakemac53, line 1: It looks like this should be rolled into the 1.1.0-dev changelog/version?

grouma:
> > We should have a changelog entry and update the user facing documentation if we haven't already.
> 
> I've updated the changelog, but I'm not user what user facing documentation needs to be updated. If everything works correctly, it should be transparent to the user.

I thought the `reportLines` was user facing in some way but I guess not. LGTM besides the issue that Jake brought up.

grouma (approved):

CHANGELOG.md
liamappelbe: What do you mean? I bumped the version because reportLines requires vm_service 7.3.0. It's possible that users have other versioning constraints keeping them on older versions of vm_service, so this needs to be a separate version of package:coverage so that pub can keep them on 1.1.0 if necessary. I ran into this issue with dwds.

grouma: `1.1.0` has not been published so we can just merge with the section below, that is `1.1.0-dev`

liamappelbe: Ah ok, I suppose that's what the -dev means. Done

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
   optional bool flag controlling whether function coverage is collected.
 * Ensure `createHitmap` returns a sorted hitmap. This fixes a potential issue with
   ignore line annotations.
+* Use the `reportLines` flag in `vm_service`'s `getSourceReport` RPC. This
+  typiclly halves the number of RPCs that the coverage collector needs to run.
```
natebosch, line 20: [nit] 
```suggestion
  typically halves the number of RPCs that the coverage collector needs to run.
```

lib/src/collect.dart
```diff
@@ -97,6 +97,8 @@ Future<Map<String, dynamic>> _getAllCoverage(
   scopedOutput ??= <String>{};
   final vm = await service.getVM();
   final allCoverage = <Map<String, dynamic>>[];
+  final ver = await service.getVersion();
```
natebosch, line 100: [nit] 
```suggestion
  final version = await service.getVersion();
```
https://dart.dev/guides/language/effective-dart/design#avoid-abbreviations

lib/src/collect.dart
```diff
@@ -97,6 +97,8 @@ Future<Map<String, dynamic>> _getAllCoverage(
...
+  final reportLines = ver.major == 3 && ver.minor != null && ver.minor! >= 51;
```
natebosch, line 101: Do we intend for this to flip off when the VM service bumps to `4.0.0` or should we add `version.major > 3 ||`?

lib/src/collect.dart
```diff
@@ -110,19 +112,22 @@ Future<Map<String, dynamic>> _getAllCoverage(
         if (!scopedOutput.contains(scope)) continue;
         final scriptReport = await service.getSourceReport(
             isolateRef.id!, <String>[SourceReportKind.kCoverage],
-            forceCompile: true, scriptId: script.id);
-        final coverage = await _getCoverageJson(
-            service, isolateRef, scriptReport, includeDart, functionCoverage);
+            forceCompile: true,
+            scriptId: script.id,
+            reportLines: reportLines ? true : null);
```
natebosch, line 117: Why does this argument take `true` or `null` and not `true` or `false`?

test/collect_coverage_api_test.dart
```diff
@@ -83,10 +83,38 @@ void main() {
...
+      return map;
+    });
+
+    for (var sampleCoverageData in sources[_sampleAppFileUri]!) {
+      expect(sampleCoverageData['funcNames'], isNotNull);
+      expect(sampleCoverageData['funcHits'], isNotNull);
+    }
+
+    for (var sampleCoverageData in sources[_isolateLibFileUri]!) {
+      expect(sampleCoverageData['funcNames'], isNotEmpty);
```
natebosch, line 108: Why do we check for non-null above and non-empty here?

natebosch (approved):

lib/src/collect.dart
liamappelbe: If the service implementer doesn't support the reportLines flag, I don't want to send it at all. The VM ignores unrecognized flags, but I'm not sure about the other backends.

lib/src/collect.dart
liamappelbe: Done

lib/src/collect.dart
liamappelbe: Done

CHANGELOG.md
liamappelbe: Done

test/collect_coverage_api_test.dart
liamappelbe: Not sure. I was just copying `test('collect_coverage_api', ...`. Switched to isNotEmpty in both tests.

PR #352, opened 2021-09-09
Gather coverage once per isolate group

liamappelbe:
Isolate groups were enabled by default in the Dart VM recently. The hit counters are shared between all isolates in the group, so this caused a regression where coverage hits were being double counted. This PR fixes that regression by only collecting coverage for the first isolate in each group. This should also give us a significant performance boost in multi-isolate programs.

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -102,8 +102,25 @@ Future<Map<String, dynamic>> _getAllCoverage(
       (version.major == 3 && version.minor != null && version.minor! >= 51) ||
           (version.major != null && version.major! > 3);
 
+  // Program counters are shared between isolates in the same group. So we need
+  // to make sure we're only gathering coverage data for one isolate in each
+  // group, otherwise we'll double count the hits.
+  final isolateOwnerGroup = <String, String>{};
+  final coveredIsolateGroups = <String>{};
+  for (var isolateGroupRef in vm.isolateGroups!) {
+    final isolateGroup = await service.getIsolateGroup(isolateGroupRef.id!);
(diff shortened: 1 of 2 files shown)


liamappelbe:
Looks like `test/collect_coverage_api_test.dart: collect_coverage_api with isolateIds` is still failing, which makes sense. This PR doesn't fix the case where one isolate from a group is filtered, but the other isn't. Since the hit counters are shared, in that case the coverage for the filtered isolate would still be collected. This isn't an issue though, because that filtering is just to save RPCs, and it doesn't actually filter the coverage numbers. Just need to fix the test to filter more isolate IDs.

grouma (approved):

grouma:
> Looks like `test/collect_coverage_api_test.dart: collect_coverage_api with isolateIds` is still failing, which makes sense. This PR doesn't fix the case where one isolate from a group is filtered, but the other isn't. Since the hit counters are shared, in that case the coverage for the filtered isolate would still be collected. This isn't an issue though, because that filtering is just to save RPCs, and it doesn't actually filter the coverage numbers. Just need to fix the test to filter more isolate IDs.

Should you include that change in this PR so that we have the full context?

liamappelbe:
> Should you include that change in this PR so that we have the full context?

Yep, I was still working on it. I tried to find a way of slightly modifying the test, but it wasn't easily fixable, so I just weakened the test instead.

natebosch:
@liamappelbe - for future PRs can you please use the "squash and merge" approach instead of a merge commit for merging PRs?

https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#handling-pull-requests

PR #355, opened 2021-12-02
Switch to a major version bump

natebosch:
We have a breaking change in public signature. Emphasize the breaking
change by making it the first bullet and making it bold.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
-## 1.1.0-dev
+## 2.0.0-dev
 
+* **BREAKING**: The signatures of `createHitmap`, `mergeHitmaps`,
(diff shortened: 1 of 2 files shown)


liamappelbe:
LGTM, but it seems there's a version solving issue. I guess test and coverage have a bit of a cyclic dependency. Do we need to update package:test as well?

natebosch:
> Do we need to update package:test as well?

Yes. We'll need to verify that `test` isn't broken and use a dependency override in the short term until we can publish. We probably want to have a full rollout plan ready before we do so that have an idea how long we'll need an override for.

natebosch:
holding off on this for now. We may want to look at making some backwards compatible upgrade routes so that we don't need atomic upgrades of multiple packages.

PR #356, opened 2021-12-03
Change function coverage to be backwards compatible

liamappelbe:
There were 5 functions that were changed in a backwards incompatible way in #348: `createHitmap`, `mergeHitmaps`, `parseCoverage`, `toScriptCoverageJson`, and `Formatter.format`.

This change makes #348 backwards compatible by renaming those newer functions, and adding versions of all those functions with the same signature as before. Most of these legacy functions simply wrap the newer versions and convert the `HitMap` back to a `Map<int, int>`.

| Legacy | New |
| --- | --- |
| createHitmap | HitMap.parseJson |
| mergeHitmaps | FileHitMaps.merge |
| parseCoverage | HitMap.parseFiles |
| toScriptCoverageJson | deprecated, no replacement |
| Formatter | deprecated, no replacement |
| PrettyPrintFormatter | FileHitMapsFormatter.prettyPrint |
| LcovFormatter | FileHitMapsFormatter.formatLcov |

### DIFF
--- a/lib/src/formatter.dart
+++ b/lib/src/formatter.dart
@@ -9,11 +9,8 @@ import 'hitmap.dart';
 
 abstract class Formatter {
   /// Returns the formatted coverage data.
-  @Deprecated('Migrate to formatV2')
+  @Deprecated('Migrate to FileHitMapsFormatter')
   Future<String> format(Map<String, Map<int, int>> hitmap);
-
-  /// Returns the formatted coverage data.
-  Future<String> formatV2(Map<String, HitMap> hitmap);
 }
 
 /// Converts the given hitmap to lcov format and appends the result to
@@ -33,56 +30,12 @@ class LcovFormatter implements Formatter {
   final String? basePath;
   final List<String>? reportOn;
 
-  @Deprecated('Migrate to formatV2')
+  @Deprecated('Migrate to FileHitMapsFormatter.formatLcov')
   @override
   Future<String> format(Map<String, Map<int, int>> hitmap) {
-    return formatV2(hitmap.map((key, value) => MapEntry(key, HitMap(value))));
-  }
-
-  @override
-  Future<String> formatV2(Map<String, HitMap> hitmap) async {
-    final pathFilter = _getPathFilter(reportOn);
-    final buf = StringBuffer();
-    for (var key in hitmap.keys) {
-      final v = hitmap[key]!;
-      final lineHits = v.lineHits;
-      final funcHits = v.funcHits;
-      final funcNames = v.funcNames;
-      var source = resolver.resolve(key);
-      if (source == null) {
-        continue;
-      }
-
-      if (!pathFilter(source)) {
-        continue;
-      }
-
-      if (basePath != null) {
-        source = p.relative(source, from: basePath);
-      }
-
-      buf.write('SF:$source\n');
-      if (funcHits != null && funcNames != null) {
-        for (var k in funcNames.keys.toList()..sort()) {
-          buf.write('FN:$k,${funcNames[k]}\n');
-        }
-        for (var k in funcHits.keys.toList()..sort()) {
-          if (funcHits[k]! != 0) {
-            buf.write('FNDA:${funcHits[k]},${funcNames[k]}\n');
-          }
-        }
-        buf.write('FNF:${funcNames.length}\n');
-        buf.write('FNH:${funcHits.values.where((v) => v > 0).length}\n');
-      }
-      for (var k in lineHits.keys.toList()..sort()) {
-        buf.write('DA:$k,${lineHits[k]}\n');
-      }
-      buf.write('LF:${lineHits.length}\n');
-      buf.write('LH:${lineHits.values.where((v) => v > 0).length}\n');
-      buf.write('end_of_record\n');
-    }
-
-    return buf.toString();
+    return Future.value(hitmap
+        .map((key, value) => MapEntry(key, HitMap(value)))
+        .formatLcov(resolver, basePath: basePath, reportOn: reportOn));
   }
 }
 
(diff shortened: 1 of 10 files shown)


kevmoo:
I'll let @natebosch run with this

lib/src/hitmap.dart
```diff
@@ -10,8 +10,12 @@ import 'package:coverage/src/util.dart';
 
 /// Contains line and function hit information for a single script.
 class HitMap {
+  /// Constructs a HitMap.
+  HitMap([Map<int, int>? _lineHits, this.funcHits, this.funcNames])
+      : lineHits = _lineHits ?? <int, int>{};
```
natebosch, line 15: The argument name does not need a leading underscore.

Also you could optionally omit the generic arguments on the map literal.

```suggestion
  HitMap([Map<int, int>? lineHits, this.funcHits, this.funcNames])
      : lineHits = lineHits ?? {};
```

lib/src/hitmap.dart
```diff
@@ -40,8 +44,27 @@ class _HitInfo {
 /// Creates a single hitmap from a raw json object. Throws away all entries that
 /// are not resolvable.
 ///
+/// DEPRECATED: Migrate to createHitmapV2.
```
natebosch, line 47: Annotated deprecated methods with `@Deprecated('migrate to createHitmapV2')`

lib/src/hitmap.dart
```diff
@@ -40,8 +44,27 @@ class _HitInfo {
...
+    checkIgnoredLines: checkIgnoredLines,
+    packagesPath: packagesPath,
+  );
+  return result.map((key, value) => MapEntry(key, value.lineHits));
+}
+
+/// Creates a single hitmap from a raw json object. Throws away all entries that
+/// are not resolvable.
+///
+/// `jsonResult` is expected to be a List<Map<String, dynamic>>.
```
natebosch, line 66: This line can be omitted, the argument type is specified in the signature

lib/src/hitmap.dart
```diff
@@ -40,8 +44,27 @@ class _HitInfo {
...
+}) async {
+  final result = await createHitmapV2(
+    jsonResult,
+    checkIgnoredLines: checkIgnoredLines,
+    packagesPath: packagesPath,
+  );
+  return result.map((key, value) => MapEntry(key, value.lineHits));
+}
+
+/// Creates a single hitmap from a raw json object. Throws away all entries that
```
natebosch, line 63: [nit] move the details to a new paragraph instead of immediately following the header sentence.

https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

lib/src/hitmap.dart
```diff
@@ -157,7 +180,29 @@ Future<Map<String, HitMap>> createHitmap(
 }
 
 /// Merges [newMap] into [result].
-void mergeHitmaps(Map<String, HitMap> newMap, Map<String, HitMap> result) {
+///
+/// DEPRECATED: Migrate to mergeHitmapsV2.
+void mergeHitmaps(
+    Map<String, Map<int, int>> newMap, Map<String, Map<int, int>> result) {
+  newMap.forEach((String file, Map<int, int> v) {
```
natebosch, line 187:
```suggestion
  newMap.forEach((file, v) {
```

https://dart.dev/guides/language/effective-dart/design#dont-annotate-inferred-parameter-types-on-function-expressions

lib/src/hitmap.dart
```diff
@@ -157,7 +180,29 @@ Future<Map<String, HitMap>> createHitmap(
...
+    final fileResult = result[file];
+    if (fileResult != null) {
+      v.forEach((int line, int cnt) {
```
natebosch, line 190:
```suggestion
      v.forEach((line, count) {
```

https://dart.dev/guides/language/effective-dart/design#avoid-abbreviations

natebosch (commented):
I worry that with the `V2` suffix the end state is going to be harder to reach. Ideally we'd like a way where the migration can be incremental, but after the final migration we'd end up with names that we are happy with.

I wonder if instead of `V2` versions of each library it might work to make a new top-level library `package:coverage/coverage_migrate.dart` with the same names and new signatures. It would still take 2 major version bumps to get to the state we want to be in:

- Version 1.x contains `/coverage.dart` with legacy and `/coverage_migrate.dart` and new. Incrementally change imports to the new version and upgrade the method calls.
- Version 2.0 drops legacy methods, moves definitions to  `/coverage.dart` and leaves `/coverage_migrate.dart` as an `export`. Incrementally change imports back without changing code.
- Version 3.0 drops the `export`.

cc @jakemac53 - WDYT? Worth an extra major version, or should we just leave in `V2` methods?

liamappelbe:
> I worry that with the `V2` suffix the end state is going to be harder to reach. Ideally we'd like a way where the migration can be incremental, but after the final migration we'd end up with names that we are happy with.
> 
> I wonder if instead of `V2` versions of each library it might work to make a new top-level library `package:coverage/coverage_migrate.dart` with the same names and new signatures. It would still take 2 major version bumps to get to the state we want to be in:
> 
> * Version 1.x contains `/coverage.dart` with legacy and `/coverage_migrate.dart` and new. Incrementally change imports to the new version and upgrade the method calls.
> * Version 2.0 drops legacy methods, moves definitions to  `/coverage.dart` and leaves `/coverage_migrate.dart` as an `export`. Incrementally change imports back without changing code.
> * Version 3.0 drops the `export`.
> 
> cc @jakemac53 - WDYT? Worth an extra major version, or should we just leave in `V2` methods?

I think in this approach we'd need a fork of at least hitmap.dart and formatter.dart. A bit of a maintenance burden, but hopefully we would bump to version 2.0 soon.

Another approach would be to come up with names for each of these functions that are less jarring than "V2". For example, `formatV2` could be `formatHitMaps`.

jakemac53:
I am probably missing some context here, but why don't we just jump right to the new apis and do a breaking change?

natebosch:
> why don't we just jump right to the new apis and do a breaking change?

Because it's going to be hard to roll atomically. We'd prefer to roll internally for validation before publishing, but we cannot start using these APIs in flutter code until they are published.

I think the alternative is to stage the internal CL for validation, but don't submit it, then publish and start rolling. We'd still need to roll this package atomically with flutter, and I don't know if we have mechanisms to do that.

natebosch:
> Another approach would be to come up with names for each of these functions that are less jarring than "V2". For example, `formatV2` could be `formatHitMaps`.

We could reconsider whether top-level methods are the right design at all for this functionality... The type `Map<String, HitMap>` could be a decent target for extension methods for some of them. Other methods might have better names available. We shouldn't pick something different just for the sake of change - but I don't know if the existing names are the most clear ones we'd choose if we were starting this library without baggage today.

Some initial proposals - a little bit of bikeshedding is welcome.

| | |
|-|-|
|`createHitmap(json)`|`HitMap.parseCoverage(json)`|
|`mergeHitmaps(newMap, result)`|`newMap.mergeWith(result)`|
|`parseCoverage(files)`|`HitMap.parseFiles(files)`|
|`toScriptCoverageJson`|maybe remove?|
|`LcovFormatter().format(hitmap)`|`hitmap.formatLcov()`|
|`PrettyPrintFormatter().format(hitMap)`|`hitmap.prettyPrint()`|

Someone copy/pasted `toScriptCoverageJson` into flutter for some reason, but I don't see other uses internally. Can we drop it from the public API? If not, can we find a more specific name? Why would someone want to use that method over the ones that deal with `Map<String, Hitmap>`?

There is no advantage to the polymorphism allowed by making `Formatter` a class, so we should probably drop it.

At the same time we can clean up some stuff like the unused second `int` argument to `parseCoverage`. We should do a review and see if there are any other design considerations worth rethinking. None of these should make the migration in packages any more difficult.

liamappelbe:
Sounds good Nate. I've implemented those changes, with some modifications:
- I changed `HitMap.parseCoverage(json)` to `HitMap.parseJson(json)` (a bit more descriptive and less confusing with the legacy `parseCoverage` function).
- For `mergeWith` I think it makes more sense for the `this` map to be the one that's modified (consistent with Map.addAll).
- For `toScriptCoverageJson`, the easiest solution it to make it a method on HitMap. I could make it private, but then I it wouldn't be accessible in collect.dart or chrome.dart.

natebosch:
> * I could make it private, but then I it wouldn't be accessible in collect.dart or chrome.dart.

For this problem I usually will make it public and top-level, but not a static in the class. Then remove it from the public API with a `show` on the `export`.

liamappelbe:
> For this problem I usually will make it public and top-level, but not a static in the class. Then remove it from the public API with a `show` on the `export`.

Done

lib/src/formatter.dart
```diff
@@ -9,7 +9,8 @@ import 'hitmap.dart';
 
 abstract class Formatter {
```
natebosch, line 10: How about deprecating this entire class in favor of a static method? Is there anywhere we are taking `Formatter` or `FileHitMapsFormatter` as an argument instead of invoking the `format` method in the same place we decide which instance to construct?

liamappelbe: Done. I don't see anywhere that the base class is referenced at all.

natebosch (approved):

PR #363, opened 2022-02-02
Updates readme to use `dart pub`over `pub`

Luckey-Elijah:
The below command for activation is deprecated and should be replace with `dart pub`

```
$ pub global activate coverage
The top level `pub` command is deprecated. Use `dart pub` instead.
```

This PR updates the README to use 

```
dart pub global activate coverage
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Tools
 
 #### Install coverage
 
-    pub global activate coverage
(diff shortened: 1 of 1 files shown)


kevmoo (approved):

kevmoo:
Thanks @Luckey-Elijah !!

PR #364, opened 2022-02-03
Handle new VM service message format

bkonyi:
"Observatory listening on..." is eventually being updated to "Dart VM
Service listening on...". This change allows for parsing the VM service
URI from messages of either format.

See https://github.com/dart-lang/sdk/issues/46756

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -50,7 +50,7 @@ Uri? extractObservatoryUri(String str) => extractVMServiceUri(str);
 /// Potentially useful as a means to extract it from log statements.
 Uri? extractVMServiceUri(String str) {
   final listeningMessageRegExp = RegExp(
-      r'(?:Observatory|Dart VM Service) listening on ((http|//)[a-zA-Z0-9:/=_\-\.\[\]]+)',
+    r'(?:Observatory|Dart VM Service) listening on ((http|//)[a-zA-Z0-9:/=_\-\.\[\]]+)',
   );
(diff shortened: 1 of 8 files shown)


lib/src/util.dart
```diff
@@ -41,17 +41,22 @@ Future<dynamic> retry(Future Function() f, Duration interval,
...
-  const kObservatoryListening = 'Observatory listening on ';
-  final msgPos = str.indexOf(kObservatoryListening);
-  if (msgPos == -1) return null;
-  final startPos = msgPos + kObservatoryListening.length;
-  final endPos = str.indexOf(RegExp(r'(\s|$)'), startPos);
-  try {
-    return Uri.parse(str.substring(startPos, endPos));
-  } on FormatException {
-    return null;
+@Deprecated('Use extractVMServiceUri instead')
```
natebosch, line 44: I don't think this is in the public API so we should be able to skip a deprecation period.

Are you aware of a `src` import with usage of this from outside the package?

natebosch (commented):
Can you add a CHANGELOG entry?

lib/src/util.dart
bkonyi: For some reason I got the impression this was used elsewhere, but it definitely isn't. I'll remove this.

bkonyi:
> Can you add a CHANGELOG entry?

Done.

natebosch (approved):

PR #366, opened 2022-02-09
Change support for VM service message from "The Dart VM Service is listening" to "The Dart VM service is listening"

bkonyi:
See https://github.com/dart-lang/sdk/issues/46756


bkonyi:
Last string change, I promise!

natebosch (approved):

PR #370, opened 2022-03-10
Deprecate --packages flag and add --package

### LINKED ISSUE #369: Deprecate --packages flag
This flag takes either the .packages or package_config.json file directly. This is a bit inconvenient, especially since .packages is being discontinued, and package_config.json is inside the .dart_tool directory, which users shouldn't have to know the details of.

Instead we should have a --package flag that accepts the path to the root of the package, and uses package_config.findPackageConfig to locate the package config.

liamappelbe:
Add a `--package` flag to format_coverage, which takes the package's root directory, instead of the .package file. This flag defaults to the current working directory, so in the common case the user won't have to pass it at all. Deprecate the `--packages` flag.

Also, deprecate the packagesPath parameter and add packagePath instead, in `HitMap.parseJson`, `HitMap.parseFiles`, `createHitmap`, and `parseCoverage`.

Fixes #369

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,3 +61,30 @@ collected. If `--sdk-root` is set, Dart SDK coverage will also be output.
 - `// coverage:ignore-line` to ignore one line.
 - `// coverage:ignore-start` and `// coverage:ignore-end` to ignore range of lines inclusive.
 - `// coverage:ignore-file` to ignore the whole file.
+
+#### Function and branch coverage
+
+To gather function level coverage information, pass `--function-coverage` to
+collect_coverage:
+
+```
+dart --pause-isolates-on-exit --disable-service-auth-codes --enable-vm-service=NNNN script.dart
+pub global run coverage:collect_coverage --uri=http://... -o coverage.json --resume-isolates --function-coverage
+```
+
+To gather branch level coverage information, pass `--branch-coverage` to *both*
+collect_coverage and the Dart command you're gathering coverage from:
+
+```
+dart --pause-isolates-on-exit --disable-service-auth-codes --enable-vm-service=NNNN --branch-coverage script.dart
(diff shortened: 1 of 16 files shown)


natebosch (commented):
In some places we've kept the `--packages` argument and use it to point to `package_config.json` instead of `.packages`.

https://pub.dev/documentation/package_config/latest/package_config.package_config/findPackageConfigUri.html

liamappelbe:
> In some places we've kept the --packages argument and use it to point to package_config.json instead of .packages.

Michael thinks we should delete the flag. We can discuss it more here: #369

README.md
```diff
@@ -44,17 +44,18 @@ all isolates are paused before collecting coverage.
 #### Formatting coverage data
 
 ```
-pub global run coverage:format_coverage --packages=app_package/.packages -i coverage.json
+pub global run coverage:format_coverage --package=app_package -i coverage.json
```
kevmoo, line 47: To pile on: should we update this to `dart pub global...`

liamappelbe: Done

liamappelbe:
Friendly ping @natebosch . I think this is ready to go. I've switched it from deleting the old flag/api to just deprecating it.

natebosch (approved):

cbracken:
@liamappelbe @natebosch looks like this was soft-breaking to Flutter (though our lint rules cause this to be hard-breaking). I've got a patch that fixes it up here:
https://github.com/flutter/flutter/pull/103771

Unfortunately, applying that patch would break google3, because package:coverage appears to not have rolled to google3 yet, as a result, we're blocked on all flutter depdendency upgrades at the moment.

Is there a plan to roll this in soon?

cbracken:
I've temporarily marked these uses as ignore to get us unblocked on the Flutter side. Filed https://github.com/flutter/flutter/issues/103830 to track.

PR #378, opened 2022-05-11
Use libraryFilters to speed up scopedOutput checks

### LINKED ISSUE #376: Use `libraryFilters` in `getSourceReport`
Once it's available, this will make the implementation of `--scope-output` much faster.

liamappelbe:
This change was only a few lines, but we didn't have any infrastructure to test it. So most of this PR is a new test that uses mockito to mock the VmService. While I was there I also tested the reportLines flow, which also wasn't able to be tested properly without mocks.

Also, had to bump the minimum SDK version to 2.15.0, because vm_service version 8.3.0 depends on it.

Fixes #376


kevmoo (approved):
:shipit:

Issue #380, opened 2022-05-12
Running `dart run coverage:test_with_coverage` yields an exception


jellynoone:
Running `dart run coverage:test_with_coverage` fails with the error being the `collect_coverage.dart` file not being found as the culprit.

Error below:
```txt
The Dart VM service is listening on http://127.0.0.1:8181/
The Dart DevTools debugger and profiler is available at: http://127.0.0.1:8181/devtools/#/?uri=ws%3A%2F%2F127.0.0.1%3A8181%2Fws
Could not find file `collect_coverage.dart`

Unhandled exception:
ProcessException: 
  Command: /usr/local/flutter/bin/cache/dart-sdk/bin/dart run collect_coverage.dart --wait-paused --resume-isolates --uri=http://127.0.0.1:8181/ --scope-output=cli_graphics -o .../coverage/coverage.json
#0      dartRun (file:///Users/.../.pub-cache/hosted/pub.dartlang.org/coverage-1.3.0/bin/test_with_coverage.dart:35:5)
<asynchronous suspension>
#1      main (file:///Users/.../.pub-cache/hosted/pub.dartlang.org/coverage-1.3.0/bin/test_with_coverage.dart:178:3)
<asynchronous suspension>
```

Additional, it seems that the source relies on finding the path for it's working dir through `Platform.script.path`, which becomes something different because `dart run package:executable` builds a snapshot of the executable and places it under a different directory: `/Users/.../package/.dart_tool/pub/bin/coverage/test_with_coverage.dart-2.17.0.snapshot`.

The culprit seems to be: https://github.com/dart-lang/coverage/blob/288f95607c4bb660154c4de2e9f7828c2375d6bf/bin/test_with_coverage.dart#L147-L149

I think a more reliable way would be to simply use the commands? Since a user must have the package installed right? The question however is how would we detect when the executable is ran on the package level and when on the global pub level? I think we could use the path information of the script to determine this.

Playing around with prints I can see two distinct `Platform.script.path` when invoking `dart pub global run ...` and `dart run ...`:
`file:///Users/.../.pub-cache/global_packages/coverage/bin/test_with_coverage.dart-2.17.0.snapshot`
and 
`file:///Users/.../.dart_tool/pub/bin/coverage/test_with_coverage.dart-2.17.0.snapshot`.

I think we could then use the path information to call the command either with `dart pub global run coverage:...` or `dart run coverage:...`.

PR #381, opened 2022-05-12
Fix running test_with_coverage non-directly

jellynoone:
Fixes: #380

Before this PR if you ran:

`dart run coverage:test_with_coverage` or `dart pub global run coverage:test_with_coverage`

you would get an exception stating  that `collect_coverage.dart` was not found.

This was because the code spawning the process was relying on `Platform.script.path` to find the package directory. This is fine if you run `dart run bin/test_with_coverage.dart` directly.

But running `dart run coverage:test_with_coverage`  or `dart pub global run coverage:test_with_coverage` in another package yields a different result, because the file is compiled to a snapshot and put in a cache directory.

From experimentation I've come to the following conclusions:
* running  `dart pub global run` produces a cache at: `~/.pub-cache/global_packages/coverage/bin/test_with_coverage.dart-2.17.0.snapshot`
* running  `dart run` at a package level produces a cache at: `path/to/package/.dart_tool/pub/bin/coverage/test_with_coverage.dart-2.17.0.snapshot`

From these observation I've modified the process running functions to detect how the file is invoked and if appropriate use the `dart ... run coverage:_command_` instead of the file directly, the reason being we the user is invoking this command they already have coverage installed somewhere, the only question is where.

While these cache directories are probably fragile to use as a means to determine what to run, they are the only way I could come up with and possibly find the right way to execute this file.

What we now do is check if the script path contains `.pub-cache` then we assume the file was ran through `dart pub global run` and then if the files isn't global we check if it contains `.dart_tool` if it doesn't we determine this was file was called directly through `dart run bin/test_with_coverage.dart`, then we use the same approach used before.

Now, there is an edge case here that a user might have the package they use under a subdirectory named `.pub-cache` but I think this is highly unlikely. Similarly as stated before relying on the path to determine what to call feels a bit of a hack, but since this command was added to replace a custom shell script I think its fine since I believe it should work for most users.

Now as for tests, I haven't added any, what makes testing it more difficult is that is tool dependent, and I also don't think there is a way to instal a package globally without it being published on pub.dev. If there is, please let me know and I'll take a look.
However, I did test this locally running all the three ways I can think of this file would be invoked by an average user:
* `dart run bin/test_with_coverage.dart`
* `dart pub global run coverage:test_with_coverage`
* `dart run coverage:test_with_coverage`

and all of them worked.

_Note: to test the package commands I had to modify the package source in the `.pub-cache` folder._


jellynoone:
Good news! I found the way to test these, and used a better way to determine which command to call and fail add a proper exception if we cannot.

Now we do the following:
- check if the file is under `.dart_tool` if it is we call `dart run coverage:*`
- check if the script's directory contains the `format_coverage.dart` and `collect_coverage.dart` files if it does, we execute them i.e. the user probably called `dart run bin/test_with_coverage.dart` directly, i.e. the package developers,
- and finally see if the path contains `global_packages`, this is used as a
- otherwise fail

To test this, we add a sub package in /test. There we essentially only depend on coverage from path.
We the run the tests by first getting the test package dependencies and globally activating the coverage package.

However, two things: 
- it is important to use the `dart` tool with a rewritten global cache from those tests, this way we can test the current version without effecting the system cache. 
- and when setting up the test for `dart pub global run` when using `dart pub global activate` we have to pass the `-s git` source not path, since path uses the cache from the package for it's executables not the actual cache a regular user would when invoking `dart pub global run`.

Now, the test scripts opens up three processes and uses three different ports, please advice if this is fine or if there are any restrictions here.

jellynoone:
Requesting review: @kevmoo, @liamappelbe, @natebosch

liamappelbe:
An even cleaner option would be to import the collector and formatter and invoke their main functions (or an internal function) directly. We really only need to start a subprocess for the tests

jellynoone:
@liamappelbe Genius!!!

I applied your second proposal, and the tests pass locally.

I also overwrote my local pub-cache with this and tried it on another project.

I still kept the new integration tests, since I think they are valuable.

test/test_with_coverage_test.dart
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
```
liamappelbe, line 1: nit: 2022

liamappelbe (commented):
Looks good so far. This will need a CHANGELOG entry, and an update to pubspec.yaml to version `1.3.1`. I'll try and publish this today, since it's an important fix.

jellynoone:
Done

liamappelbe (approved):

PR #394, opened 2022-06-15
test: drop special cases for Dart v1 SDKs

kevmoo:


kevmoo:
This will continue to fail, but it shouldn't fail any worse. This is just...junk.

I did test @liamappelbe SDK fix at SDK HEAD. Still one failure here, but it's unrelated.

liamappelbe:
> I did test @liamappelbe SDK fix at SDK HEAD. Still one failure here, but it's unrelated.

Yeah, there's was a small branch coverage change caused by my fix in the VM. I just need to update the test when the dev branch rolls.

liamappelbe (approved):

PR #400, opened 2022-06-16
Ignore implicit functions

liamappelbe:
Also add a proper function coverage test.


kevmoo (approved):
It's tough to see what would be skipped here. Could you include an example?

liamappelbe:
> It's tough to see what would be skipped here. Could you include an example?

Added an example to the CL description

natebosch (approved):

PR #402, opened 2022-06-23
Fix handling of part files in function coverage

### LINKED ISSUE #399: Function coverage can't handle part files
When `_processFunction` is called for a function defined in a part file, it searches for the token pos to line mapping in the root file of the library.

Related to #389

liamappelbe:
The current logic looks up the script's library, then looks up all the functions in the library, and assumes that all those functions are in the script. But part files break this assumption.

So instead we have to look up the library, then for each of its functions, look up the corresponding script, and insert the function into that script's hitmap. Caching means that this isn't any slower.

Fixes #399

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -256,24 +256,29 @@ int? _getLineFromTokenPos(Script script, int tokenPos) {
 }
 
 Future<void> _processFunction(VmService service, IsolateRef isolateRef,
-    Script script, FuncRef funcRef, HitMap hits) async {
+    Future<Script?> Function(ScriptRef?) scriptGetter, FuncRef funcRef, HitMap hits) async {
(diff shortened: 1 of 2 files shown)


lib/src/collect.dart
```diff
@@ -271,22 +306,12 @@ Future<void> _processFunction(VmService service, IsolateRef isolateRef,
       }
       return;
     }
+    final hits = getHitMap(Uri.parse(script.uri!));
+    hits.funcHits ??= <int, int>{};
+    hits.funcNames ??= <int, String>{};
```
kevmoo, line 311: nit. You *could* merge these lines.

```suggestion
    (hits.funcNames ??= <int, String>{})[line] = funcName;
```

🤷  – get rid of a null assert.

kevmoo (approved):

PR #412, opened 2022-07-06
Ignore abstract methods

### LINKED ISSUE #398: Function coverage considers abstract methods missed
Abstract methods should be ignored.

liamappelbe:
Fixes #398

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -11,6 +11,7 @@ import 'util.dart';
 import 'hitmap.dart';
 
 const _retryInterval = Duration(milliseconds: 200);
+const _debugTokenPositions = bool.fromEnvironment('DEBUG_COVERAGE');
 
 /// Collects coverage for all isolates in the running VM.
 ///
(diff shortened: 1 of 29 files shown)


lib/src/collect.dart
```diff
@@ -284,7 +284,7 @@ Future<List<Map<String, dynamic>>> _getCoverageJson(
 
   Future<void> processFunction(FuncRef funcRef) async {
     final func = await service.getObject(isolateRef.id!, funcRef.id!) as Func;
-    if (func.implicit ?? false) {
+    if ((func.implicit ?? false) || (func.isAbstract ?? false)) {
```
kevmoo, line 287: Maybe? 🤷 

```suggestion
    if (func.implicit ?? func.isAbstract ?? false) {
```

liamappelbe: This would change the result if implicit is false but abstract is true. Old code would say true, but this change would say false.

kevmoo: Ah! Yup...

kevmoo:
Looks like we'll have to drop Dart 2.15?

liamappelbe:
> Looks like we'll have to drop Dart 2.15?

No, it's fine. Just need to make that test conditional. We already have a few tests that do different things on different versions.

kevmoo (approved):

PR #414, opened 2022-08-11
ignore-line comments etc must be at the end of the line

### LINKED ISSUE #413: File gets removed from coverage wrongly when "// coverage:ignore-file" is used in Strings
## Steps to Reproduce

1. Create new dart project using `dart create  example`
2. Update `lib/example.dart` so it looks like this:

```dart
int calculate() {
  print('// coverage:ignore-file'); // <- this line leads to the file being excluded from coverage
  return 6 * 7;
}
```
3. Execute `dart test --coverage` from the root of the project

**Expected results:** 
The generated `coverage/lcov.info` should contain coverage information about the file `lib/example.dart`

**Actual results:** 
The generated `coverage/lcov.info` does NOT contain  coverage information about the file `lib/example.dart`

liamappelbe:
This fixes #413, without the hassle of actually parsing all the strings (downside of this simple fix is that it's still possible for this bug to happen in multi-line strings). Before this, `print('// coverage:ignore-file');` would disable coverage for the file.

The fix is basically to make sure the line ends with the comment, rather than just containing it. But if I just used String.endsWith, then you'd end up with some very hard to debug errors:

```dart
// One of these is invalid:
// coverage:ignore-file
// coverage:ignore-file     
```

So instead I'm using regexps like `r'// coverage:ignore-file\s*$'`, to allow trailing white-space.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.5.1
+
+- Fix bug where 'ingore-line' comments etc are applied even if they're inside
+  string literals.
+
 ## 1.5.0
 
 - Support passing extra arguments to `test_with_coverage` which are then passed
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -74,10 +74,10 @@ Future<int> getOpenPort() async {
   }
 }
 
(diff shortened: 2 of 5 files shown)


kevmoo (approved):
:shipit:

PR #419, opened 2022-08-18
Add a benchmark

### LINKED ISSUE #418: Add a benchmark to github CI
I'm planning to migrate flutter test from a custom coverage implementation to package:coverage. The main issue with this is performance, since package:coverage is slower than their simplified implementation. The performance gap will be closed soon, but we should add a simple benchmark to detect if we slow things down again.

The pathological case for coverage performance is when we have many small tests all running in separate isolates, so the benchmark should be something like that.

liamappelbe:
Fixes #418


liamappelbe:
Seems like it's not a specific benchmark that's slow, but rather that each of the benchmarks can flakily hang. I'll set up timeout and retry logic.

liamappelbe:
> Hrm...see that things are still running after 2 hours?

Interesting. All the benchmarks finished in a reasonable amount of time, but it's hanging afterwards. Maybe some of the isolates are failing to shut down. I'll make this a draft again, until I figure this out.

.github/workflows/test-package.yml
```diff
@@ -77,3 +77,31 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           path-to-lcov: coverage/lcov.info
+
+  benchmark:
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
kevmoo, line 85: In general, we've been pushing towards pinning actions at the latest version released version with a SHA – FYI

kevmoo (approved):

PR #425, opened 2022-10-10
Fix void_checks lint that happens with latest analysis

kevmoo:


natebosch:
Hmm, I would expect botg `() => throw ''` and `() { throw '';}` to have a static and runtime type of `Never Function()`. It's not clear to me why we'd need to make this change.

@bwilkerson - is there a regression in the linter which caused this, or does this look like a legitimate `void_checks` lint?

bwilkerson:
The most recent changes to this rule were innocuous: changing some getters in analyzer that had been deprecated as part of a rename and fixing a typo in the docs. Other than that there hasn't been a change since last year.

bwilkerson:
I think we infer a static type in the arrow case, but not in the block case. @stereotype441 to confirm.

natebosch:
> I think we infer a static type in the arrow case, but not in the block case.

Is this a recent change in the analyzer?

This doesn't match the behavior when there is no context type. In both `var x = () => throw '';` and `var x = () { throw ''; };` the static type from the analysis server, and runtime type on the VM are all `() => Never`.

When passing either `() => throw '';` or `() { throw '';}` to an argument with type `FutureOr<void> Function()` the runtime type is `() => Never`. I'm not sure the static type in that case.

kevmoo:
Thanks for doing the follow-up here, @natebosch !!

bwilkerson:
> Is this a recent change in the analyzer?

As far as I know there hasn't been any recent change. I just thought that's what I remembered, but I don't trust my memory on the semantics, so I pinged Paul.

kevmoo:
Are we okay to land this? @natebosch Should we have a tracking issue open on the analyzer?

natebosch (approved):

natebosch:
Ah, I think I was confused about what was changing. This lint _would have_ always shown up, and we're fixing it now because it was just recently enabled?

I'm confused by the semantics, but not as worried as I would have been if this was a previously enabled lint that started surfacing in a new SDK version.

stereotype441:
I just dug into the analyzer and linter logic on this, and I agree that this is not a recent change.  The linter specifically suppresses the lint in the case where the expression takes the form `() => ...`.

Personally I think the lint could be improved so that it doesn't fire in this case.  I'll file a linter issue.

stereotype441:
> I just dug into the analyzer and linter logic on this, and I agree that this is not a recent change. The linter specifically suppresses the lint in the case where the expression takes the form `() => ...`.
> 
> Personally I think the lint could be improved so that it doesn't fire in this case. I'll file a linter issue.

Filed https://github.com/dart-lang/linter/issues/3766.

PR #464, opened 2023-10-02
Prepare for 1.6.4 release

bkonyi:
Includes upgrade to vm_service 12.0.0. This change needs to be published to unblock rolling the new version of vm_service through to Flutter.


jakemac53 (approved):
LGTM, just remove the overrides now so the stable jobs can select an older version

PR #466, opened 2023-10-17
Add `coverableLineCache` param to `collect`

liamappelbe:
`coverableLineCache` is a map from library URI to the set of all lines that are coverable. If users are doing multiple `collect` calls over the same set of libraries, they should pass an empty map to this param in the first call, and then pass the same map to all subsequent calls.

If `coverableLineCache` is provided, we set `VmService.getSourceReport`'s `librariesAlreadyCompiled` param to all its keys. This disables forced compilation for each of these libraries. The effect of this is that no misses are returned in the response for these libraries (and coverage collection is faster). To fill in the missing misses, we use the values of the `coverableLineCache` map. Finally, we fill this cache with all the hit and missed lines we've seen for each library.

So `coverableLineCache` caches all the lines that have been hit or missed in previous `collect` calls, and uses it to skip recompiling those libraries, and to fill out the misses of subsequent calls.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -4,7 +4,7 @@ description: Coverage data manipulation and formatting
 repository: https://github.com/dart-lang/coverage
 
 environment:
-  sdk: '>=3.0.0 <4.0.0'
+  sdk: '^3.0.0'
 
 dependencies:
   args: ^2.0.0
@@ -13,7 +13,7 @@ dependencies:
   path: ^1.8.0
   source_maps: ^0.10.10
   stack_trace: ^1.10.0
-  vm_service: '>=12.0.0 <13.0.0'
+  vm_service: '^12.0.0'
 
 dev_dependencies:
   benchmark_harness: ^2.2.0
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: coverage
-version: 1.6.4
+version: 1.7.0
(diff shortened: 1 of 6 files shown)


pubspec.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   path: ^1.8.0
   source_maps: ^0.10.10
   stack_trace: ^1.10.0
-  vm_service: '>=11.9.0 <13.0.0'
+  vm_service: '>=12.0.0 <13.0.0'
```
bkonyi, line 16: Nit: If we're constraining vm_service to a single major revision, we should just use `^12.0.0` instead of an explicit range.

liamappelbe:
> Is there any particular reason we're keeping track of the coverable lines for each script instead of just keeping track of which scripts have already been compiled?

The purpose of force compiling all the libraries is that it tells the VM which lines are coverable, so it can generate a complete list of all the missed lines (particularly for functions that haven't been called). Each collection that the flutter tester does is in a totally fresh VM, so the VM doesn't have any knowledge about which lines are coverable left over from previous collections. So when we set the `librariesAlreadyCompiled` flag, the libraries that are not force compiled won't have this data, so there will be a bunch of misses that are not included in the source report. To fill in the missing misses, we cache all the coverable lines from previous runs.

bkonyi (approved):

PR #494, opened 2024-07-08
update lints

kevmoo:


devoncarew:
@liamappelbe, @bkonyi - we're still seeing https://github.com/dart-lang/coverage/issues/490 (failures on windows). If this is a broader issue (due to plumbing changes around `dds`?), we should file an issue in the sdk repo to track.

devoncarew (approved):

Issue #427, opened 2022-11-09
Filter files using a command line flag


liamappelbe:
Add a command line flag that works like `// coverage:ignore-file`. It should also be able to handle globs, so that users can ignore all generated files like `*.g.dart`.

devoncarew:
I wonder if we should glob `*.g.dart` by default? Is there value in getting coverage info for generated code? Would excluding it by default be unexpected / surprising for users?

They would have the cli flag to disable the behavior if it was undesired.

liamappelbe:
I think I would find that surprising, if I was a user. In package:wasm I'd want to disable that behavior. I think the distinction is between users who have generated code that was generated by a 3rd party tool (eg protos), or generated by a tool that is part of the package (which is the case in package:wasm). The former case is probably more common, but in the latter case changing this default behavior would be a confusing breaking change.

As for whether we just add a flag to exclude `*.g.dart`, or add a flag that excludes arbitrary globs, I think it's probably going to be the same amount of work either way, so I'm inclined to just do the more general arbitrary glob approach.

natebosch:
I'm in favor of a flag which allows arbitrary globs.

osa1:
How widely use used is `*.g.dart` format for generated files? We don't use it in protoc_plugin, we use `*.pb.dart`.

I agree with @liamappelbe that whether I want to skip generated files or not depends on the use case.

1. If I'm using a code generator (e.g. protoc_plugin) then I want to skip generated files. Testing of generated code should be done by the code generator.
2. If I'm developing a code generator then I want coverage for generated files as well as I'm testing correctness of those in my package.

In general there will be more of (1) than (2) so ignoring generated files might make sense, but we should make sure code generators use a common suffix like `.g.dart` for generated files. Currently protoc_plugin does not do that (uses `.pb.dart`). If not then the default will be useless for some of (2) as well.

Either way we should have a flag for arbitrary globs.

If we ignore some files by default then the design of flags gets a bit more complicated as we will probably also need a flag to not ignore anything by default, or have a flag that is opposite of ignore-file (maybe allow-file).

If we don't ignore by default then one flag for ignoring files will do.

I say we don't ignore by default to keep things simple for now. If we want to ignore some files by default in the future it will be the same breaking change when we add it now.

iapicca:
it would be nice to have a yaml file `coverage_options.yaml`
(like [lints](https://pub.dev/packages/lints) uses `analysis_options.yaml`)
[...]

Issue #313, opened 2020-09-28
Ignore files from coverage by path pattern


shouichi:
Currently, `// coverage:ignore-file` is available for ignoring the whole file from coverage but some tools (such as built_value_generator) do not generate files with that meta-comment. It would be nice if we can ignore generated files from coverage by specifying something like `--exclude *.g.dart`. We could fix tools to add the meta-comment but it may make more sense to let developers choose whether to exclude generated files from coverage or not. Thank you.

cbracken:
There are currently a variety of flags relating to scoping coverage collection/reporting.

collect_coverage has:
* `--scope-output`: limit coverage to just the specified package path
* `--include-dart`: compute coverage for dart: libraries

format_coverage has:
* `report-on`: specify files/directories to report coverage on

I wonder if it might make sense to add a matching `--no-report-on` option to `format_coverage` where people can list a set of glob patterns to exclude? Would that work for you?

Side note to myself/other maintainers: long term, I think it'd be nice for us to move these options into `collect_coverage` and reduce the amount of input earlier in the process. The tricky part there is file path to package URI mapping, but this is probably significantly easier than it was before Dart 2.0.

shouichi:
> Would that work for you?

We use flutter and it emits `lcov.info` file when running `flutter test --coverage`. I honestly don't know the internal but it seems like flutter doesn't use `format_coverage`.

https://github.com/flutter/flutter/blob/e8fbb43aacb70568e6c3d37bcad4f6f36e490523/packages/flutter_tools/lib/src/test/coverage_collector.dart#L116

It seems like flutter will support excluding certain files from coverage. It might be better to wait and see what happens. (Ignoring files from coverage is not critical for us.)

https://github.com/flutter/flutter/issues/25077#issuecomment-699538134

Thank you.

cbracken:
It's been a long time since I originally added coverage support to Flutter, but I know most of the original code has been replaced with something a bit more custom and performant by @jonahwilliams. 

It might be worth filing a [new issue](https://github.com/flutter/flutter/issues/new/choose) against Flutter to request the ability to filter files for collection (`--coverage-include` and `--coverage-exclude` multi-flag that takes glob patterns) requesting this feature.

[...]

PR #496, opened 2024-07-29
Add --ignore-files option to exclude files from code coverage using path patterns

gogolon:
This makes it possible to exclude files which match glob patterns from the report, which was discussed in #427 and #313. The `.yaml` configuration part which was requested in #427 has not been implemented, but it can be done separately after this is merged.

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.


bin/format_coverage.dart
```diff
@@ -124,50 +127,54 @@ Future<void> main(List<String> arguments) async {
 Environment parseArgs(List<String> arguments) {
   final parser = ArgParser();
 
-  parser.addOption('sdk-root', abbr: 's', help: 'path to the SDK root');
```
kevmoo, line 127: careful making unrelated changes. Makes reviews harder.

bin/format_coverage.dart
```diff
@@ -66,6 +68,7 @@ Future<void> main(List<String> arguments) async {
   final hitmap = await HitMap.parseFiles(
     files,
     checkIgnoredLines: env.checkIgnore,
+    ignoreGlobs: env.ignoreFiles,
```
liamappelbe, line 71: Why not add this logic in the same place as `reportOn`? You could just add it to the `_getPathFilter` function in formatter.dart. Having the logic in two totally different places makes it a bit hard to read.

test/lcov_test.dart
```diff
@@ -122,6 +124,37 @@ void main() {
...
+    test('formatLcov() excludes files matching glob patterns', () async {
+      final hitmap = await _getHitMap();
+
+      final resolver = await Resolver.create(packagePath: '.');
+      final res = hitmap.formatLcov(
+        resolver,
+        ignoreGlobs: {Glob('**/*.g.dart'), Glob('**/util.dart')},
+      );
+
+      expect(res, isNot(contains(p.absolute(_sampleGeneratedPath))));
```
liamappelbe, line 136: I think you should be able to add an extra check for this file to all the other tests in this file: `expect(res, contains(p.absolute(_sampleGeneratedPath)));`. That way you're checking both the contains and doesn't contain cases.

At the moment I think these tests could pass accidentally by, for example, mistyping the `_sampleGeneratedPath`, since every test that checks it is checking that it's not in the report.

liamappelbe (commented):
Looking good. Just need to tighten up the tests a bit.

test/lcov_test.dart
gogolon: I added this check as well as some additional checks in globs related tests to make the expected behaviour more explicit

liamappelbe (approved):

PR #498, opened 2024-07-30
Apply `--scope-output` filters after `getSourceReport`

### LINKED ISSUE #495: --scope-output does not work
Sometimes coverage is gathered from external packages even when `--scope-output` flag is used and set to the name of the app's package.

**Reproduction**
1. Create a new dart project with `dart create minimal-sample`
2. Add equatable package: `dart pub add equatable`
3. Replace the code inside `lib/minimal_sample.dart` with the code below:
```
import 'package:equatable/equatable.dart';

class SomeClass with EquatableMixin {
  @override
  List<Object?> get props => [];
}

int calculate() {
  return 6 * 7;
}
```
4. run 
```
dart run --pause-isolates-on-exit --disable-service-auth-codes --enable-vm-service=8181 test &
dart pub global run coverage:collect_coverage --wait-paused --uri=http://127.0.0.1:8181/ -o coverage/coverage.json --resume-isolates --scope-output=minimal_sample
```

After performing the steps above, the generated `coverage/coverage.json` looks like this:

```
{
  "type": "CodeCoverage",
  "coverage": [
    {
      "source": "package:minimal_sample/minimal_sample.dart",
      "script": {
        "type": "@Script",
        "fixedId": true,
        "id": "libraries/1/scripts/package%3Aminimal_sample%2Fminimal_sample.dart",
        "uri": "package:minimal_sample/minimal_sample.dart",
        "_kind": "library"
      },
      "hits": [
        4,
        0,
        5,
        0,
        8,
        1,
        9,
        1
      ]
    },
    {
      "source": "package:equatable/src/equatable_mixin.dart",
      "script": {
        "type": "@Script",
        "fixedId": true,
        "id": "libraries/1/scripts/package%3Aequatable%2Fsrc%2Fequatable_mixin.dart",
        "uri": "package:equatable/src/equatable_mixin.dart",
        "_kind": "library"
      },
      "hits": [
        16,
        0,
        18,
        0,
        21,
        0,
        22,
        0,
        23,
        0,
        26,
        0,
        27,
        0,
        29,
        0,
        31,
        0,
        32,
        0,
        33,
        0,
        34,
        0,
        35,
        0,
        37,
        0,
        38,
        0,
        39,
        0
      ]
    }
  ]
}
```

I use Dart SDK 3.3.3.

liamappelbe:
The `--scope-output` filters are passed to `getSourceReport.libraryFilters`, but there are cases where the report can contain ranges that bypass the filter. This happens when the range has a different library than their enclosing function/class (eg mixins).

To fix this we just need to re-apply the filters to the reported ranges. It's still important to pass the filters to `getSourceReport`, because these are pretty rare edge cases and the filtering is still a useful optimisation to reduce the size of the report.

Fixes #495

### DIFF
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -507,10 +507,15 @@ class StdoutLog extends Log {
 extension _ScopedOutput on Set<String> {
   bool includesScript(String? scriptUriString) {
     if (scriptUriString == null) return false;
+
+    // If the set is empty, it means the user didn't specify a --scope-output
+    // flag, so allow everything.
     if (isEmpty) return true;
+
     final scriptUri = Uri.parse(scriptUriString);
     if (scriptUri.scheme != 'package') return false;
-    final scope = scriptUri.path.split('/').first;
+
+    final scope = scriptUri.pathSegments.first;
     return contains(scope);
   }
 }
--- a/lib/src/collect.dart
+++ b/lib/src/collect.dart
@@ -178,11 +178,8 @@ Future<Map<String, dynamic>> _getAllCoverage(
         continue;
       }
       for (final script in scripts.scripts!) {
-        final uri = Uri.parse(script.uri!);
-        if (uri.scheme != 'package') continue;
-        final scope = uri.path.split('/').first;
         // Skip scripts which should not be included in the report.
-        if (!scopedOutput.contains(scope)) continue;
+        if (!scopedOutput.includesScript(script.uri)) continue;
         late final SourceReport scriptReport;
         try {
           scriptReport = await service.getSourceReport(
@@ -203,7 +200,8 @@ Future<Map<String, dynamic>> _getAllCoverage(
             includeDart,
             functionCoverage,
             reportLines,
-            coverableLineCache);
+            coverableLineCache,
+            scopedOutput);
         allCoverage.addAll(coverage);
       }
     } else {
@@ -229,7 +227,8 @@ Future<Map<String, dynamic>> _getAllCoverage(
           includeDart,
           functionCoverage,
           reportLines,
-          coverableLineCache);
+          coverableLineCache,
+          scopedOutput);
       allCoverage.addAll(coverage);
     }
   }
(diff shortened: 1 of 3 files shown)


lib/src/collect.dart
```diff
@@ -497,3 +503,14 @@ class StdoutLog extends Log {
...
   void severe(String message) => print(message);
 }
+
+extension _ScopedOutput on Set<String> {
+  bool includesScript(String? scriptUriString) {
+    if (scriptUriString == null) return false;
+    if (isEmpty) return true;
+    final scriptUri = Uri.parse(scriptUriString);
+    if (scriptUri.scheme != 'package') return false;
+    final scope = scriptUri.path.split('/').first;
```
bkonyi, line 513: Nit: would `final scope = scriptUri.pathSegments.first` achieve the same thing?

lib/src/collect.dart
```diff
@@ -497,3 +503,14 @@ class StdoutLog extends Log {
   @override
   void severe(String message) => print(message);
 }
+
+extension _ScopedOutput on Set<String> {
+  bool includesScript(String? scriptUriString) {
+    if (scriptUriString == null) return false;
+    if (isEmpty) return true;
```
bkonyi, line 510: Shouldn't this return false if the set is empty? If not, can you leave a comment explaining why we're considering an empty set to include all scripts?

test/collect_coverage_mock_test.dart
```diff
@@ -628,5 +628,52 @@ void main() {
...
+
+    test(
+        'Collect coverage, scoped output, library filters, '
+        'handles SourceReports that contain unfiltered ranges', () async {
+      // Regression test for https://github.com/dart-lang/coverage/issues/495
+      final service = _mockService(3, 57);
+      when(service.getSourceReport('isolate', ['Coverage'],
+              forceCompile: true,
+              reportLines: true,
+              libraryFilters: ['package:foo/']))
```
bkonyi, line 640: Ubernit: consider adding trailing commas for formatting.

lib/src/collect.dart
liamappelbe: Added a comment

lib/src/collect.dart
liamappelbe: Done

test/collect_coverage_mock_test.dart
liamappelbe: Done

lib/src/collect.dart
bkonyi: Thanks! That makes sense.

bkonyi (approved):

